PR #43, opened 2018-03-21
Add login support for sovereigns

yujhongmicrosoft:


yujhongmicrosoft:
@smichelotti

chrmarti:
@yujhongmicrosoft Thanks for your contribution!

Looking through the code, I see two general issues:
- We store the refresh token using `keytar`, but we currently don't remember to which cloud that refresh token belongs to, so the automatic login with an existing refresh token for any sovereign cloud will fail because the code will try to use the refresh token with the Azure Public cloud. Could you store which environment the refresh token belongs to using `keytar` too?
- The selected environment is forgotten after a restart of VS Code, could you store that in a setting?

yujhongmicrosoft:
@chrmarti I **think** the changes I made above have resolved the two issues you pointed out:
       - I set the credentialsService to the name of the environment, so refresh tokens will be stored with 
         the correct environment. 
       - I saved the current environment using keytar, so that when the user restarts VS Code the 
         environment will be set to what it was last. If this is the first time the user is using this extension, the 
         selected environment will be set to Public. 

Please let me know what other changes need to be made! 
Will be more quick about it this time :) 

Thank you!

package.nls.json
```diff
@@ -1,10 +1,11 @@
...
+	"azure-account.commands.setEnvironment": "Set Azure Environment",
 	"azure-account.commands.logout": "Sign Out",
 	"azure-account.commands.selectSubscriptions": "Select Subscriptions",
 	"azure-account.commands.createAccount": "Create an Account",
 	"azure-account.commands.openCloudConsoleLinux": "Open Bash in Cloud Shell",
-	"azure-account.commands.openCloudConsoleWindows": "Open PowerShell in Cloud Shell",
-	"azure-account.commands.uploadFileCloudConsole": "Upload to Cloud Shell"
-}
\ No newline at end of file
+	"azure-account.commands.openCloudConsoleWindows": "Open PowerShell in Cloud Shell"
```
chrmarti: Did a comma go missing here?

src/azure-account.ts
```diff
@@ -598,8 +646,8 @@ export async function tokenFromRefreshToken(refreshToken: string, tenantId = com
 async function tokensFromToken(firstTokenResponse: TokenResponse) {
 	const tokenCache = new MemoryCache();
 	await addTokenToCache(tokenCache, firstTokenResponse);
-	const credentials = new DeviceTokenCredentials({ username: firstTokenResponse.userId, clientId, tokenCache });
-	const client = new SubscriptionClient(credentials);
+	const credentials = new DeviceTokenCredentials({ environment: selectedEnvironment, username: firstTokenResponse.userId, clientId, tokenCache });
+	const client = new SubscriptionClient(credentials, selectedEnvironment.resourceManagerEndpointUrl);
```
chrmarti: It looks like all our code will need to pass the correct baseUrl when creating a client. That makes the change breaking for dependent extensions. It would make sense to first ship the feature in a hidden way, such that dependent extensions can update and test it.

To do that: Please remove the 'Set Environment' command and instead read from a hidden (i.e., not declared in package.json) setting `azure.experimental.cloud` which cloud to use when signing in.

src/azure-account.ts
```diff
@@ -444,7 +492,7 @@ export class AzureLoginHelper {
 	private async loadSubscriptions() {
 		const lists = await Promise.all(this.api.sessions.map(session => {
 			const credentials = session.credentials;
-			const client = new SubscriptionClient(credentials);
+			const client = new SubscriptionClient(credentials, selectedEnvironment.resourceManagerEndpointUrl);
```
chrmarti: cloudConsole.ts uses a TenantDetailsClient that might need a similar parameter.

src/azure-account.ts
```diff
@@ -25,6 +25,7 @@ import TelemetryReporter from 'vscode-extension-telemetry';
 const localize = nls.loadMessageBundle();
 
 const keytar = getNodeModule<typeof keytarType>('keytar');
+var selectedEnvironment : AzureEnvironment;
```
chrmarti: Instead of using a global variable, there should usually be an AzureSession object in any context where an environment is needed. Please use that. (That will keep the code more flexible towards multiple concurrent logins.)

chrmarti: With that you can then move reading and storing of the environment in keytar to the same places where the refresh token is read and stored.

src/azure-account.ts
```diff
@@ -40,16 +41,48 @@ function getNodeModule<T>(moduleName: string): T | undefined {
...
+if(keytar.findPassword("storedEnvironment") == null){
+	keytar.setPassword("storedEnvironment", "current", "Azure");
+}
+
+//set selectedenvironment to current saved in keytar
+var storedenv = keytar.getPassword("storedEnvironment", "current");
+storedenv.then(function(env) {
+    selectedEnvironment = getEnvironment(env);
+});
+vscode.window.setStatusBarMessage("Current Azure Environment: " + selectedEnvironment.name);
```
chrmarti: For now only set the status bar message when it is not the public cloud. (See other comment on hiding the feature for now.)

src/azure-account.ts
```diff
@@ -40,16 +41,48 @@ function getNodeModule<T>(moduleName: string): T | undefined {
...
-const defaultEnvironment: AzureEnvironment = (<any>AzureEnvironment).Azure;
 const commonTenantId = 'common';
-const authorityHostUrl = defaultEnvironment.activeDirectoryEndpointUrl; // Testing: 'https://login.windows-ppe.net/'
 const clientId = 'aebc6443-996d-45c2-90f0-388ff96faa56'; // VSC: 'aebc6443-996d-45c2-90f0-388ff96faa56'
 const validateAuthority = true;
-const authorityUrl = `${authorityHostUrl}${commonTenantId}`;
-const resource = defaultEnvironment.activeDirectoryResourceId;
 
-const credentialsService = 'VSCode Public Azure';
+const credentialsService = selectedEnvironment.name;
```
chrmarti: Lets keep that 'VSCode Public Azure'. It might have been a mistake on my part to have 'Public' in the name there. This way we only need to manage one refresh token.

src/azure-account.ts
```diff
@@ -294,7 +342,7 @@ export class AzureLoginHelper {
 		const cache: Cache = {
 			subscriptions: this.api.subscriptions.map(({ session, subscription }) => ({
 				session: {
-					environment: session.environment.name,
+					environment: <any>(selectedEnvironment),
```
chrmarti: `session.environment.name` should still be correct.

kingliantop:
Hi, which version we can use to login the national cloud? The latest release 0.42 still dosen't provide the option. Thanks!

chrmarti:
@kingliantop This is work in progress and not available yet.

chrmarti:
Continued in #79.

PR #46, opened 2018-03-26
add timeout in promise race

jdneo:
If the user click `copy & open` but doing nothing else. Then the extension will keep waiting for the response. So I added a timeout here.

The 3 minutes is just estimated by myself.


chrmarti:
Merged with modifications. Thanks!

Issue #45, opened 2018-03-26
Could we continue the command 'azure-account.askForLogin' when creating cloud shell


jdneo:
Hi @chrmarti 

Just find that when we are creating the cloud shell, and the user is not logged in, the extension will execute the code below: https://github.com/Microsoft/vscode-azure-account/blob/master/src/cloudConsole.ts#L125

Then if the user is not logged in previously, he needs to trigger the command again.

Could we await this command instead of return here? 

for example:
```javascript
...
if (!(await api.waitForLogin())) {
    await commands.executeCommand("azure-account.askForLogin");
    if (!(await api.waitForLogin())) {
        // handle login failure
    }
}
...
```

jdneo:
Hi @chrmarti , I sent out a PR about this issue: https://github.com/Microsoft/vscode-azure-account/pull/48. 😄

chrmarti:
I have so far taken the view that when a user wants to do something (e.g., Open Cloud Shell) that requires the user to sign in first, it makes sense to let the user repeat the initial action (Open Cloud Shell) because there is such a long time between taking that initial action and completing the sign in process. The user might simply have forgotten what s/he was up to by the time the sign in completes.

Could you describe the scenario you are looking at?

jdneo:
Yep, the scenario is just like you said. 

1. User clicks "Open Cloud Shell".
2. User is told that he needs to sign in first.
3. User signs in.
4. Cloud Shell is provisioned.
5. Cloud Shell is shown to the user.

I agree that this is not a problem if users just forget their intention after signing in. But what if the user still remember? He might keep waiting but nothing happens.

I think continue the execution might improve the experience with the latter uses, as long as this change won't introduce regression. 😄

jdneo:
Em, actually this feedback is from our testers. After they click open cloud shell, nothing happens after signing in, which makes them confused.

BTW, an idea just came up in my mind.

Maybe I can check the login status through the API, and if the user is not signed in, tell him that he needs to sign in first, and then trigger the login command from the account extension.

After login, next time user trigger cloud shell related command, he can pass the login status check and do whatever he wants to do.

But I still much prefer the one click experience, since less action is needed.

What do you think?

jdneo:
@chrmarti I saw your code change and tested on my side. I like the experience that telling user sign in needed in the terminal panel. Thanks!

But when user hasn't signed in at begining, I got Entry not found in cache. error when calling acquireToken(). Is that becaus I missed something?

[...]

PR #48, opened 2018-03-26
change return to await when provision cloud shell

jdneo:
resolve: #45


jdneo:
Thanks @chrmarti . Looking forward to the new experience 😄

jdneo:
@chrmarti I saw your code change and tested on my side. I like the experience that telling user sign in needed in the terminal panel. Thanks!

PR #81, opened 2018-08-06
add azure environment configuration

Sneezry:


chrmarti:
Thanks. We are waiting for some testing and downstream changes to support the other clouds.

forester123:
Hi, what's the status of this? We have some docs of Azure China reply on this update.

chrmarti:
We are working on enabling our app id for Azure China and we will need dependent extensions to update to support this before we can add the setting for real.

ArthurMa1978:
@chrmarti can you give us a estimation of when will this pr be merged and new release within this change? We have been asked to support mooncake (China region) for long time from local customer.

chrmarti:
Fixing with #83.

chrmarti:
@ArthurMa1978 Should be early this week.

ArthurMa1978:
cool, thanks @chrmarti !

PR #94, opened 2018-10-02
Update vscode dependency

ramya-rao-a:
The current version of `vscode` has dependencies that have known security vulnerabilities


chrmarti:
Thanks!

PR #95, opened 2018-10-02
Commands translation for italian

HCRenegade:
Added command translation for Italian.

git


chrmarti:
Hi @HCRenegade,

Thanks for your contribution! Unfortunately we currently don't have a story for translating all display strings to a particular language. Including your translations would only translate a small part of the extension and would be outdated as soon as we add more commands. I'm therefore not merging it.

Thanks for your understanding.

PR #143, opened 2019-06-21
Use the new `@azure/arm` and `@azure/ms-rest` APIs

PrashanthCorp:
** This PR is not backward compatible **

Updated the repo to use the new `@azure/arm` and `@azure/ms-rest` APIs.


PrashanthCorp:
To test it, clone the PR branches: 
https://github.com/microsoft/vscode-azuretools/pull/511
https://github.com/microsoft/vscode-cosmosdb/pull/1121
This branch

Run `npm i` in each repo
- In azuretools/ui, run `npm run build`, then  `npm link`
- In azure-account run `vsce package`
- Make sure the cosmos repo references the local copy of azure-tools
- For azure-account, install the locally built vsix. 

Now run the cosmos extension in debug mode.

src/azure-account.api.d.ts
```diff
@@ -4,10 +4,10 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { Event, Terminal, Progress, CancellationToken } from 'vscode';
-import { ServiceClientCredentials } from 'ms-rest';
-import { AzureEnvironment } from 'ms-rest-azure';
-import { SubscriptionModels } from 'azure-arm-resource';
+import { ServiceClientCredentials } from '@azure/ms-rest-js';
+import { Environment } from '@azure/ms-rest-azure-env';
 import { ReadStream } from 'fs';
+import { SubscriptionModels } from '@azure/arm-subscriptions';
```
chrmarti: I wonder if the types we actually use are compatible between new and old SDKs. If so, could we inline the interfaces here and not import any of the new (nor old) SDK's types?

src/azure-account.ts
```diff
@@ -460,7 +463,7 @@ export class AzureLoginHelper {
 			environment,
 			userId: tokenResponse.userId!,
 			tenantId: tokenResponse.tenantId!,
-			credentials: new DeviceTokenCredentials({ environment: environment, username: tokenResponse.userId, clientId, tokenCache: this.delayedCache, domain: tokenResponse.tenantId })
+			credentials: new DeviceTokenCredentials(clientId, tokenResponse.tenantId, tokenResponse.userId, undefined, defaultEnvironment, this.tokenCache)
```
chrmarti: Why pass `defaultEnvironment` and not the `environment`?

PrashanthCorp: Nice catch.

src/azure-account.api.d.ts
PrashanthCorp: From what I can tell, they don't seem compatible. There are a couple of fields a few layers under these types that are incompatible. I remember one case had an issue with the way the signRequest method was defined (callback with void vs an awaitable Promise<T>). 

From what I understand there are small changes in the typings (callback -> thenable, string -> enum, etc.)

In this case, using the `SubscriptionModels` from `azure-arm-resource` leads to an error like: 
```
 Types of property 'subscription' are incompatible.
      Type 'import("c:/repo/vscode-azure-account/node_modules/azure-arm-resource/lib/subscription/models/index").Subscription' is not assignable to type 'import("c:/repo/vscode-azure-account/node_modules/@azure/arm-subscriptions/esm/models/index").Subscription'.
        Types of property 'state' are incompatible.
          Type 'string | undefined' is not assignable to type '"Enabled" | "Warned" | "PastDue" | "Disabled" | "Deleted" | undefined'.
            Type 'string' is not assignable to type '"Enabled" | "Warned" | "PastDue" | "Disabled" | "Deleted" | undefined'.ts(2322)
```

package.json
```diff
@@ -222,16 +222,23 @@
   },
   "dependencies": {
     "adal-node": "0.1.28",
-    "azure-arm-resource": "7.0.1",
+    "copy-paste": "^1.3.0",
+    "ms-rest-azure": "2.5.9",
+    "vscode-nls": "4.0.0",
+    "opn": "^5.1.0",
```
EricJizbaMSFT: Looks like you added some extra dependencies on accident. These are unrelated to `@azure/` packages.

src/azure-account.api.d.ts
```diff
@@ -4,10 +4,10 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { Event, Terminal, Progress, CancellationToken } from 'vscode';
-import { ServiceClientCredentials } from 'ms-rest';
-import { AzureEnvironment } from 'ms-rest-azure';
-import { SubscriptionModels } from 'azure-arm-resource';
+import { ServiceClientCredentials } from '@azure/ms-rest-js';
```
EricJizbaMSFT: Old:
```
export interface ServiceClientCredentials {
  /**
   * Signs a request with the Authentication header.
   *
   * @param {WebResource} The WebResource to be signed.
   * @param {function(error)}  callback  The callback function.
   */
  signRequest(webResource: WebResource, callback: { (err: Error): void }): void;
}
```

New:
```
export interface ServiceClientCredentials {
  /**
   * Signs a request with the Authentication header.
   *
   * @param {WebResource} webResource The WebResource/request to be signed.
   * @returns {Promise<WebResource>} The signed request object;
   */
  signRequest(webResource: WebResource): Promise<WebResource>;
}
```

Proposed inline type:
```
export interface ServiceClientCredentials {
  /**
   * Signs a request with the Authentication header.
   *
   * @param {WebResource} webResource The WebResource/request to be signed.
   * @returns {Promise<WebResource>} The signed request object;
   */
  signRequest(webResource: WebResource): Promise<WebResource>;

  /**
   * Signs a request with the Authentication header.
   * @deprecated use other signRequest instead
   *
   * @param {WebResource} The WebResource to be signed.
   * @param {function(error)}  callback  The callback function.
   */
  signRequest(webResource: WebResource, callback: { (err: Error): void }): void;
}
```

You'll have to modify the instance of `ServiceClientCredentials` to support the callback, but it shouldn't be that hard to convert a promise to a callback.

src/azure-account.api.d.ts
```diff
@@ -27,7 +27,7 @@ export interface AzureAccount {
 }
 
 export interface AzureSession {
-	readonly environment: AzureEnvironment;
+	readonly environment: Environment;
```
EricJizbaMSFT: I abbreviated this quite a bit, but you get the idea:

```
export declare class Environment {
    static readonly AzureCloud: {
    };

    /**
     * @deprecated use `AzureCloud` instead
     */
    static readonly Azure: {
    };


    static readonly ChinaCloud: {
    };

    /**
     * @deprecated use `ChinaCloud` instead
     */
    static readonly AzureChina: {
    };
}
```

You'll just have to copy the `AzureCloud` property over to an `Azure` property.

src/azure-account.api.d.ts
EricJizbaMSFT: `string | undefined` is not assignable to `"Enabled" | "Warned" | "PastDue" | "Disabled" | "Deleted" | undefined`, but it _will_ work the other way around.

src/azure-account.api.d.ts
chrmarti: I agree, that could work. If it gets too messy, we could consider adding a property `credentials2` with the new instance alongside `credentials` with the old one.

PR #151, opened 2019-09-16
Adopt URI handler based auth flow

RMacfarlane:
Registers a URI handler and uses the `createAppUri` API to complete auth instead of spinning up a local server.

I wasn't able to test these changes end to end, as the allowed redirect urls need to be updated for the oauth app we're using. For local selfhosting in the web, the redirect will be `http://localhost:9888`

For VSO, the endpoints are
```
https://online.visualstudio.com

https://online-ppe.core.vsengsaas.visualstudio.com

https://online.dev.core.vsengsaas.visualstudio.com
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,11 +12,12 @@
     "color": "#0072c6",
     "theme": "dark"
   },
+  "enableProposedApi": true,
   "version": "0.8.4",
   "aiKey": "AIF-d9b70cd4-b9f9-4d70-929b-a071c400b217",
   "publisher": "ms-vscode",
   "engines": {
-    "vscode": "^1.31.0"
+    "vscode": "^1.38.0"
   },
   "categories": [
     "Azure"
--- a/src/codeFlowLogin.ts
+++ b/src/codeFlowLogin.ts
@@ -108,12 +108,15 @@ async function exchangeCodeForToken(clientId: string, environment: AzureEnvironm
 
 async function loginWithoutLocalServer(clientId: string, environment: AzureEnvironment, adfs: boolean, tenantId: string): Promise<TokenResponse> {
 	const callbackUri = await vscode.env.createAppUri();
-	const query = parseQuery(callbackUri);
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -12,11 +12,12 @@
     "color": "#0072c6",
     "theme": "dark"
   },
+  "enableProposedApi": true,
```
chrmarti: Will require allow-listing the extension id in the product.json.

src/codeFlowLogin.ts
```diff
@@ -61,7 +62,72 @@ export async function checkRedirectServer(adfs: boolean) {
...
+	}).catch(err => {
+		uriEventListener.dispose();
+		throw err;
+	});
+}
+
+async function loginWithoutLocalServer(clientId: string, environment: AzureEnvironment, adfs: boolean, tenantId: string): Promise<TokenResponse> {
+	const callbackUri = (await vscode.env.createAppUri()).toString();
+	const nonce = crypto.randomBytes(16).toString('base64');
+	const signInUrl = `${environment.activeDirectoryEndpointUrl}${adfs ? '' : `${tenantId}/`}oauth2/authorize?response_type=code&client_id=${encodeURIComponent(clientId)}&redirect_uri=${encodeURIComponent(callbackUri)}&state=${nonce}&resource=${encodeURIComponent(environment.activeDirectoryResourceId)}&prompt=select_account`;
```
chrmarti: When the callbackUri is `http://localhost:port/path` you can use `redirectUrlAAD` and encode the port in the state parameter like below. That server will then redirect to localhost. (This helps with the fact that you cannot register a `http` URL as a redirect URL, not even for localhost.)

chrmarti: For ADFS just use the callbackUri for now. Each ADFS instance has its own allow-list of redirect URIs and that includes the port number whereas for AAD the hostname is enough. So ADFS might not work in the web for now. I think we should wait for someone to have the need for this to work first before investing time.

package.json
RMacfarlane: Have added it there 👍

PR #196, opened 2020-06-10
Add support for configuring credentialsSection

### LINKED ISSUE #42: Allow being signed in to multiple accounts at the same time
Request from a customer of the Azure App Service extension:

> can the extension be made to support multiple subs cause its annoying to switch every time.
[Microsoft/vscode-azureappservice/issues/319]

phealy:
This pull request adds a configuration variable called `azure.credentialsSection`. It can be used to store Azure account configuration in a different section in the system keystore, which will allow for workspace level configuration of which Azure credentials are being used.

This was the lowest impact way I could think of to resolve my particular issue, which was making use of multiple VS Code "profiles" by starting it with different `--extensions-dir` and `--user-data-dir` arguments. Because the credentials were being stored using `keytar` in the system credential store, my Azure account was following me between profiles when I didn't want it to.

Fixes #42


phealy:
Just for further details - this is what it looks like in Windows Credential Manager when this new patch is used. Now I can set a user-level setting (which follows my --user-data-dir) to change this value for each profile, and thus be logged into a different Azure account in each one. The same thing would work between workspaces if you set the property at the workspace level, and because all I changed was a previously-hardcoded value, all the existing support for multiple clouds wasn't impacted. In addition, I defaulted the value to be the same as the previous const string, so it should be a seamless transition for users - if they don't set this value, nothing will be different.

image

wwlorey:
Apologies for not getting to this sooner, @phealy. In https://github.com/microsoft/vscode-azure-account/pull/329 we removed our direct dependency on keytar and started relying on [VS Code's SecretStorage API](https://code.visualstudio.com/api/references/vscode-api#SecretStorage). This means that VS Code now sets the cache key that the extension uses to store secrets.

If this is still a feature that you'd like to see, please file an issue and we can discuss possible workarounds/solutions. Closing since this PR is no longer applicable.

PR #241, opened 2021-01-13
Get environments for azure stack

v-tsun:


src/azure-account.ts
```diff
@@ -773,17 +785,47 @@ async function getEnvironments(): Promise<Environment[]> {
...
-		return [
-			...staticEnvironments,
-			{
-				...ppe,
-				name: azurePPE
-			}
-		]
-	} else {
-		return staticEnvironments;
+	const apiProfile = config.get<Boolean>('target_azurestack_api_profile');
```
bganapa, line 788: I prefer adding code comments here,

src/azure-account.ts
```diff
@@ -773,17 +785,47 @@ async function getEnvironments(): Promise<Environment[]> {
...
+	try {
+		if (ppe) {
+			const activeDirectoryUrl = ppe.activeDirectoryEndpointUrl.endsWith('/') ? ppe.activeDirectoryEndpointUrl.slice(0,-1) : ppe.activeDirectoryEndpointUrl;
+			const validateAuthority = activeDirectoryUrl.endsWith('/adfs') ? false : true;
+			if (apiProfile) {
+				let resourceMangerUrl = ppe.resourceManagerEndpointUrl;
```
RMacfarlane, line 795: small typo - should be `resourceManagerUrl`

src/azure-account.ts
v-tsun: @bganapa What do think about this one

src/azure-account.ts
```diff
@@ -773,17 +785,47 @@ async function getEnvironments(): Promise<Environment[]> {
...
-			...staticEnvironments,
-			{
-				...ppe,
-				name: azurePPE
-			}
-		]
-	} else {
-		return staticEnvironments;
+	const apiProfile = config.get<Boolean>('target_azurestack_api_profile');
+	try {
```
RMacfarlane, line 790: I know I suggested using a try-catch here, but since it just re-throws the error, I think it isn't needed. Since you check this `target_azurestack_api_profile` property first, others who are using the `ppe` setting for testing won't be impacted

src/azure-account.ts
```diff
@@ -773,17 +785,47 @@ async function getEnvironments(): Promise<Environment[]> {
...
+							portalUrl: ppeMetadata.portalEndpoint,
+							galleryEndpointUrl: ppeMetadata.galleryEndpoint,
+							activeDirectoryGraphResourceId: ppeMetadata.graphEndpoint,
+							storageEndpointSuffix: resourceMangerUrl.substring(resourceMangerUrl.indexOf('.')),
+							keyVaultDnsSuffix: '.vault'.concat(resourceMangerUrl.substring(resourceMangerUrl.indexOf('.'))),
+							managementEndpointUrl: ppeMetadata.authentication.audiences[0],
+							validateAuthority: validateAuthority
+						}]
+				}
+				return [
```
RMacfarlane, line 816: We also need to return this if `apiProfile` is false

src/azure-account.ts
```diff
@@ -773,17 +785,47 @@ async function getEnvironments(): Promise<Environment[]> {
...
-	} else {
-		return staticEnvironments;
+	const apiProfile = config.get<Boolean>('target_azurestack_api_profile');
+	try {
+		if (ppe) {
+			const activeDirectoryUrl = ppe.activeDirectoryEndpointUrl.endsWith('/') ? ppe.activeDirectoryEndpointUrl.slice(0,-1) : ppe.activeDirectoryEndpointUrl;
+			const validateAuthority = activeDirectoryUrl.endsWith('/adfs') ? false : true;
+			if (apiProfile) {
+				let resourceMangerUrl = ppe.resourceManagerEndpointUrl;
+				resourceMangerUrl = resourceMangerUrl.endsWith('/') ? resourceMangerUrl.slice(0,-1) : resourceMangerUrl;
```
RMacfarlane, line 796: To make it a little bit clearer why we're doing this, I think it would be good to extract everything inside this `if` block to a new function called something like `getAzureStackEnvironments`

src/azure-account.ts
```diff
@@ -134,6 +134,18 @@ interface ICloudMetadata {
 	gallery: string;
 }
 
+interface IPPEMetaData {
```
bganapa, line 137: pls rename this as IResourceManagerMetadata

src/azure-account.ts
bganapa: @RMacfarlane what do we want to return if apiProfile is false?

bganapa: I feel like the if(apiProfile) check is not needed at all here, as the response we are returning should be applicable for all other custom PPE as well, just that we will be using the older api version to get the metadata

v-tsun:
@bganapa

src/azure-account.ts
RMacfarlane: Azure Stack is not the only thing that can change the PPE setting, if others are using it to test something, we need to preserver the original behavior and return the list of environments as before

src/azure-account.ts
```diff
@@ -773,19 +785,88 @@ async function getEnvironments(): Promise<Environment[]> {
...
+		return await getPpeEnvironments(ppe, apiProfile);
+	}
+	return staticEnvironments;
+}
+
+async function getPpeEnvironments(ppe: Environment, apiProfile: boolean|undefined): Promise<Environment[]> {
+	// get validateAuthority from activeDirectoryUrl from user setting, it should be set to false only under ADFS environemnt.
+	const activeDirectoryUrl = ppe.activeDirectoryEndpointUrl.endsWith('/') ? ppe.activeDirectoryEndpointUrl.slice(0,-1) : ppe.activeDirectoryEndpointUrl;
+	const validateAuthority = activeDirectoryUrl.endsWith('/adfs') ? false : true;
+	let resourceManagerUrl = ppe.resourceManagerEndpointUrl;
```
RMacfarlane, line 800: These can be declared `const` and can be moved into the `if` since they are only used in that case

src/azure-account.ts
```diff
@@ -773,19 +785,88 @@ async function getEnvironments(): Promise<Environment[]> {
...
+	let endpointsUrl = getMetaDataEndpoints(resourceManagerUrl, apiProfile);
+	const ppeResponse = await fetch(endpointsUrl);
+	if (ppeResponse.ok) {
+		if (apiProfile) {
+			// get azure stack environments
+			return await getAzureStackEnvironments(ppe, ppeResponse, validateAuthority, resourceManagerUrl);
```
RMacfarlane, line 806: I would actually move all of the fetch logic into this function

src/azure-account.ts
```diff
@@ -773,19 +785,88 @@ async function getEnvironments(): Promise<Environment[]> {
...
+			activeDirectoryGraphResourceId: ppeMetadata.graphEndpoint,
+			storageEndpointSuffix: resourceManagerUrl.substring(resourceManagerUrl.indexOf('.')),
+			keyVaultDnsSuffix: '.vault'.concat(resourceManagerUrl.substring(resourceManagerUrl.indexOf('.'))),
+			managementEndpointUrl: ppeMetadata.authentication.audiences[0],
+			validateAuthority: validateAuthority
+		}
+	]
+}
+
+function getMetaDataEndpoints(resourceManagerUrl: string, apiProfile: boolean|undefined): string {
```
RMacfarlane, line 862: `Metadata` instead of `MetaData`

src/azure-account.ts
```diff
@@ -134,6 +134,18 @@ interface ICloudMetadata {
 	gallery: string;
 }
 
+interface IResourceManagerMetaData {
```
RMacfarlane, line 137: `Metadata` here as well

RMacfarlane (changes requested):

v-tsun:
@bganapa

src/azure-account.ts
```diff
@@ -773,17 +785,71 @@ async function getEnvironments(): Promise<Environment[]> {
 
 	const config = workspace.getConfiguration('azure');
 	const ppe = config.get<Environment>('ppe');
+	// get api profile from user setting, this needs to be true for running azure stack
+	const apiProfile = config.get<boolean>('target_azurestack_api_profile');
```
RMacfarlane, line 789: I think this should be moved into the `getPpeEnvironments` function since it is only needed there

RMacfarlane (commented):
This looks good! I only have one very minor comment and then I think this is ready to be merged! Thank you!

bganapa (approved):

bganapa:
This should be good to merge!. Thanks @RMacfarlane !

PR #247, opened 2021-02-09
Add api profile to azure session environment

v-tsun:


src/accountEnvironmentsList.ts
```diff
@@ -0,0 +1,7 @@
+import { Environment } from "@azure/ms-rest-azure-env";
+import { AzureAccountEnvironment } from "./azure-account.api";
+
+export const accountAzureCloud: AzureAccountEnvironment = {...Environment.AzureCloud, azureStackApiProfile: false};
```
bganapa, line 4: I believe it is just an addition of extra property, therefore it is not breaking change for the existing consumers of this API

bganapa (approved):

src/accountEnvironmentsList.ts
v-tsun: Make api profile an optional property

src/azure-account.api.d.ts
```diff
@@ -27,8 +27,12 @@ export interface AzureAccount {
 	createCloudShell(os: 'Linux' | 'Windows'): CloudShell;
 }
 
+export interface AzureAccountEnvironment extends Environment {
+	azureStackApiProfile?: boolean;
```
RMacfarlane, line 31: What is an Api Profile? I think `isAzureStack` is clearer to me, but I'm not sure if I'm missing additional context

src/codeFlowLogin.ts
```diff
@@ -14,12 +14,13 @@ import * as net from 'net';
...
 import { Environment } from '@azure/ms-rest-azure-env';
 import { TokenResponse, AuthenticationContext } from 'adal-node';
+import { AzureAccountEnvironment } from './azure-account.api';
 
 export const redirectUrlAAD = 'https://vscode-redirect.azurewebsites.net/';
 const portADFS = 19472;
 const redirectUrlADFS = `http://127.0.0.1:${portADFS}/callback`;
 
-export function isADFS(environment: Environment) {
+export function isADFS(environment: AzureAccountEnvironment) {
```
RMacfarlane, line 23: Since the environment itself now has this flag, can you remove this function and use environment.isAzureStack or environment.azureStackApiProfile in its place?

src/codeFlowLogin.ts
```diff
@@ -150,7 +151,7 @@ async function loginWithoutLocalServer(clientId: string, environment: Environmen
 	return Promise.race([exchangeCodeForToken(clientId, environment, tenantId, callback, state), timeoutPromise]);
 }
 
-export async function login(clientId: string, environment: Environment, adfs: boolean, tenantId: string, openUri: (url: string) => Promise<void>, redirectTimeout: () => Promise<void>) {
+export async function login(clientId: string, environment: AzureAccountEnvironment, adfs: boolean, tenantId: string, openUri: (url: string) => Promise<void>, redirectTimeout: () => Promise<void>) {
```
RMacfarlane, line 154: for the functions where we pass `adfs` as a boolean but have the environment, we should get rid of the `adfs` parameter

RMacfarlane (changes requested):
Looks good overall, just a couple comments

src/codeFlowLogin.ts
v-tsun: Hey Rachel, I am not sure if we can get rid of this. Since Azure Stack can be either AAD or ADFS. I believe we still need ADFS check for us. In https://github.com/microsoft/vscode-azure-account/blob/4f1a49ec5fe4f7dec21b93bc736cf2f3af61cb3e/src/codeFlowLogin.ts#L189. I believe we will miss this if we don't do ADFS check.

src/azure-account.api.d.ts
v-tsun: "azure.target_azurestack_api_profile": true is the parameter we set in user setting. That is why Bala wants to use it as the name. @bganapa, do you think isAzureStack would be better for us. Or you still want to keep this one. If so, maybe you can jump in to give a more detailed explaination to Rachel about your thought.

bganapa: azureStackApiProfile is more appropriate here. APi Profile is a map of supported Resource Types and api versions.
More details here.
https://docs.microsoft.com/en-us/azure-stack/user/azure-stack-version-profiles?view=azs-2008
Our customers access all the tools and SDKs through API Profile. For example, in Azure CLI, to work against the azurestack, the customer does the following
az cloud update --profile 2019-03-01-hybrid
Note that there can be multiple AuzreStack API Profiles supported at the same time. In the last few years , we have only one API profile active(supported). So we decided to keep it simple for VScode and made it boolean. 

Also the customer can work against AzureCloud or MoonCake with the same AzureStack API Profile. For example, if a customer wants to develop an App that works in AzureStack as well as Azure, we recommend using the API Profile SDKs instead of using two different SDKs (one for Azure and another for AzureStack)

Hope that naming makes sense

src/codeFlowLogin.ts
bganapa: Yes, AzureStack environment can be configured either with AAD (connected) or ADFS (disconnected)

RMacfarlane (approved):
Thanks Martin and Bala for the explanations! I think this is good to merge then.

PR #251, opened 2021-02-22
Fix redirection, update account api and rename to CustomCloud

v-tsun:


package.json
```diff
@@ -127,21 +127,21 @@
             "AzureChinaCloud",
             "AzureGermanCloud",
             "AzureUSGovernment",
-            "AzureStack"
+            "CustomCloud"
```
bganapa, line 130: AzureCustomCloud

package.json
```diff
@@ -127,21 +127,21 @@
...
           ],
           "enumDescriptions": [
             "Azure",
             "Azure China",
             "Azure Germany",
             "Azure US Government",
-            "Azure Stack"
+            "Custom Cloud"
```
bganapa, line 137: Azure Custom Cloud

src/azure-account.api.d.ts
```diff
@@ -27,12 +27,8 @@ export interface AzureAccount {
 	createCloudShell(os: 'Linux' | 'Windows'): CloudShell;
 }
 
-export interface AzureAccountEnvironment extends Environment {
-	// No need for "azureStackApiProfile" if we just check Environment.name === 'AzureStack'
-}
-
 export interface AzureSession {
-	readonly environment: AzureAccountEnvironment;
+	readonly environment: Environment;
```
bganapa, line 31: I think this needs to be AzureAccountEnvironment to have the apiprofile boolean

package.json
```diff
@@ -127,21 +127,21 @@
             "AzureChinaCloud",
```
bganapa, line 127: please prefer to cheery pick the comit and make a PR directly to microsoft/vscode-azure-account instead of Erics branch

bganapa (changes requested):
few comments

src/azure-account.api.d.ts
v-tsun: Do we still need apiProfile property? Since I believe what Eric was thinking was we can decide if it is azure stack through the name of environment. If so, I am not sure we still need to leave apiProfile here.

bganapa: Yes, we still need it, otherwise we need to call the cloud exactly as AzureStack instead of AzureCustomCloud

If we choose to call it as AzureStack , in future when other restricted clouds and Azure Stack Edge clouds  needs support, we need to create a new cloud naming. I believe Custom Cloud is generic and we should have Apiprofile property exposed

v-tsun: I see, so still a boolean value? Btw, I believe even through we add this property, looks like it is an property which is not using in the lower level extensions. I am not sure if Eric would prefer us to have this kind of value which is not using in other extensions.

ejizba: > Yes, we still need it, otherwise we need to call the cloud exactly as AzureStack instead of AzureCustomCloud

> If we choose to call it as AzureStack , in future when other restricted clouds and Azure Stack Edge clouds needs support, we need to create a new cloud naming. I believe Custom Cloud is generic and we should have Apiprofile property exposed

Can you expand on this? I like what @v-tsun was doing so far in this regard. How would Azure Stack Edge and other restricted clouds be different? Depending on the answer to that question, I could see us adding new settings or properties in the future, rather than having a new name each time.

package.json
ejizba: Yeah no need to PR onto my branch

src/azure-account.api.d.ts
bganapa: @ejizba APIProfile and the SDK that we are using are just for AzureStack.  The restricted clouds may be supporting completely a different set of API Versions. They are likely to aligned closer to Azure rather than AzureStack

AzureStack Edge may be supporting different set of api versions as well and it is likely to be aligned closer to AzureStack

ejizba: I'm skeptical a boolean would be sufficient at that point anyways. I would rather ignore that scenario for now or go all in on an actual "apiProfile" setting. We can limit it to just two supported values and it would be essentially the same thing as your boolean, but more flexible for the future:
```json
"azure.apiProfile": {
  "type": "string",
  "enum": [
    "latest",
    "2019-03-01-hybrid"
  ],
  "enumDescriptions": [
    "The most recent API versions released in global Azure",
    ""
  ],
  "default": "latest",
  "description": "The API version for Azure REST endpoints."
},
```
If we're trying to get this done faster, though, I think we could assume all CustomClouds are "2019-03-01-hybrid" for now and add support for the setting later once we need to handle the other clouds.

bganapa: Agree, a boolean may not be enough. lets go ahead as you proposed as we are trying to get this one faster

v-tsun:
Make another PR https://github.com/microsoft/vscode-azure-account/pull/252 to master branch, close this one

v-tsun:
Reopen this PR since decide to remove apiProfile property and leave the name as AzureCustomCloud

v-tsun:
Close since this is merging onto Eric's branch

PR #252, opened 2021-02-24
Fix redirection, rename to CustomCloud and update apiProfile in API

v-tsun:


src/azure-account.ts
```diff
@@ -783,71 +811,70 @@ async function getEnvironments(): Promise<AzureAccountEnvironment[]> {
...
+					activeDirectoryEndpointUrl: endpoints.authentication.loginEndpoint.endsWith('/') ? endpoints.authentication.loginEndpoint : endpoints.authentication.loginEndpoint.concat('/'),
+					activeDirectoryGraphResourceId: endpoints.graphEndpoint,
+					activeDirectoryResourceId: endpoints.authentication.audiences[0],
+					portalUrl: endpoints.portalEndpoint,
+					galleryEndpointUrl: endpoints.galleryEndpoint,
+					storageEndpointSuffix: armUrl.substring(armUrl.indexOf('.')),
+					keyVaultDnsSuffix: '.vault'.concat(armUrl.substring(armUrl.indexOf('.'))),
+					managementEndpointUrl: endpoints.authentication.audiences[0],
+					validateAuthority: getValidateAuthority(endpoints.authentication.loginEndpoint),
+					apiProfile: getApiProfile(),
```
ejizba, line 854: I thought @bganapa said we should ignore apiProfile for now.

I said [this](https://github.com/microsoft/vscode-azure-account/pull/251#discussion_r581293657):

> If we're trying to get this done faster, though, I think we could assume all CustomClouds are "2019-03-01-hybrid" for now and add support for the setting _later_ once we need to handle the other clouds.

He said this:

> Agree, a boolean may not be enough. lets go ahead as you proposed as we are trying to get this one faster

ejizba: If we _do_ want to do the setting, here are some things I think we need to figure out:

- I'm not sure the `apiProfile` setting fits well on `Environment`. Theoretically I could switch api profile regardless of cloud (Azure, Azure China, Custom, etc.) and at any time, which doesn't work well with the existing events like `onSessionsChanged`.
- We'd have to figure out how to tell people to switch _back_ to "latest" if they switch from a custom cloud to Azure global. Wouldn't want them stuck on the older profile.

We can figure those out, but it would be faster to just ignore apiProfile until we actually _need_ the setting

bganapa: @v-tsun as we discussed, lets remove the apiProfile setting and assume that if AzureCustomCloud, it is for AzureStack.

We can even go for explicit naming it as AzureStack instead of AzureCustomCloud? @ejizba

ejizba: I could go either way on the naming, but Storage Explorer uses "Custom", so I lean towards that
Screen Shot 2021-02-24 at 11 49 55 AM

v-tsun: > I could go either way on the naming, but Storage Explorer uses "Custom", so I lean towards that
> Screen Shot 2021-02-24 at 11 49 55 AM

Hey Eric, so you prefer to name it as AzureCustomCloud right? so @bganapa, how about we keep our name as AzureCustomCloud?

v-tsun:
Close this PR because we decide to ignore apiProfile and name as AzureCustomCloud right now, create PR https://github.com/microsoft/vscode-azure-account/pull/253 to merge onto master branch

v-tsun:
Close this PR because we decide to ignore apiProfile and name as AzureCustomCloud right now, create PR https://github.com/microsoft/vscode-azure-account/pull/253 to merge onto master branch

PR #253, opened 2021-02-25
Fix redirection, update account API and revert AzureAccountEnvironment

v-tsun:


v-tsun:
Hey @ejizba, @bganapa, this is the newest PR for azure account. Could you guys take a look on it.

bganapa:
@v-tsun  there is a conflicting file, pls fix

src/azure-account.ts
```diff
@@ -364,7 +368,7 @@ export class AzureLoginHelper {
 
 	async loginToCloud(): Promise<void> {
 		const current = await getSelectedEnvironment();
-		const selected = await window.showQuickPick<{ label: string, description?: string, environment: AzureAccountEnvironment }>(getEnvironments()
+		const selected = await window.showQuickPick<{ label: string, description?: string, environment: Environment }>(getEnvironments(true /* includePartial */)
 			.then(environments => environments.map(environment => ({
```
bganapa, line 372: What does includePartial mean here? why do we need this flag?

src/azure-account.ts
```diff
@@ -783,71 +802,64 @@ async function getEnvironments(): Promise<AzureAccountEnvironment[]> {
...
+					portalUrl: endpoints.portalEndpoint,
+					galleryEndpointUrl: endpoints.galleryEndpoint,
+					storageEndpointSuffix: armUrl.substring(armUrl.indexOf('.')),
+					keyVaultDnsSuffix: '.vault'.concat(armUrl.substring(armUrl.indexOf('.'))),
+					managementEndpointUrl: endpoints.authentication.audiences[0],
+					validateAuthority: getValidateAuthority(endpoints.authentication.loginEndpoint)
+				};
 			}
-		]
+		} catch {
```
bganapa, line 847: why do we ignore here?

src/azure-account.ts
v-tsun: https://github.com/microsoft/vscode-azure-account/blob/2b89a79aa2127bce49c2448ff28b500c86eb6a92/src/azure-account.ts#L825

bganapa:
@ejizba @RMacfarlane  Could you pls review?

src/azure-account.ts
```diff
@@ -783,81 +802,68 @@ async function getEnvironments(): Promise<AzureAccountEnvironment[]> {
...
-			...staticEnvironments,
-			{
-				...ppe,
-				name: azurePPE,
-				validateAuthority: validateAuthority
+
+	return result;
+}
+
+async function getCustomCloudEnvironment(config: WorkspaceConfiguration, includePartial: boolean): Promise<Environment | undefined> {
```
bganapa, line 825: includePartial - need a better naming here

bganapa: we may not need this extra boolean, you can return the azurestack specific endpoints, if the cloud is set as AzureCustomCloud

v-tsun: I don't think we can remove it, if we remove it we need to have another flag. I believe for this boolean, Eric wants to get all possible environment for the window picker including custom cloud. But he doesn't want to fetch endpoints every time to avoid the traffic. And we cannot just get it from user setting. If so, then users will miss the option of custom cloud in picker. But Eric wants user can use Custom Cloud even through they haven't set cloud as CustomCloud in user setting when they pick Custom cloud in window picker. As follows, https://github.com/microsoft/vscode-azure-account/blob/0411207adf53abb5baca06044fb412a48c2293ac/src/azure-account.ts#L381-L387 Probably we can give it a better name than remove this flag. What do you think about includeCustomCloud or something?

src/azure-account.ts
```diff
@@ -783,81 +802,68 @@ async function getEnvironments(): Promise<AzureAccountEnvironment[]> {
...
+					galleryEndpointUrl: endpoints.galleryEndpoint,
+					storageEndpointSuffix: armUrl.substring(armUrl.indexOf('.')),
+					keyVaultDnsSuffix: '.vault'.concat(armUrl.substring(armUrl.indexOf('.'))),
+					managementEndpointUrl: endpoints.authentication.audiences[0],
+					validateAuthority: getValidateAuthority(endpoints.authentication.loginEndpoint)
+				};
 			}
-		]
+		} catch {
+			// ignore
```
RMacfarlane, line 848: Let's show a notification to the user if this fails, so that they can potentially take action to fix the endpoints url

src/azure-account.ts
```diff
@@ -376,9 +380,21 @@ export class AzureLoginHelper {
 		if (selected) {
 			const config = workspace.getConfiguration('azure');
 			if (config.get('cloud') !== selected.environment.name) {
+				if (selected.environment.name === azureCustomCloud) {
+					const armUrl = await window.showInputBox({
+						prompt: localize('azure-account.enterArmUrl', "Enter the Azure Resource Manager endpoint"),
+						placeHolder: 'https://management.local.azurestack.external'
+					});
+					if (armUrl) {
+						await config.update(customCloudArmUrlKey, armUrl, getCurrentTarget(config.inspect(customCloudArmUrlKey)));
```
RMacfarlane, line 389: There is an `onDidChangeConfiguration` listener earlier in this file that need to check if this key has changed

src/azure-account.ts
```diff
@@ -376,9 +380,21 @@ export class AzureLoginHelper {
 		if (selected) {
 			const config = workspace.getConfiguration('azure');
 			if (config.get('cloud') !== selected.environment.name) {
+				if (selected.environment.name === azureCustomCloud) {
+					const armUrl = await window.showInputBox({
+						prompt: localize('azure-account.enterArmUrl', "Enter the Azure Resource Manager endpoint"),
```
RMacfarlane, line 385: I suggest adding `ignoreFocusOut: true` here as well, to avoid losing the prompt easily while the user finds the url they need

src/azure-account.ts
RMacfarlane: `this.doLogin = true;` also needs to be set here, before the update

RMacfarlane (changes requested):

src/azure-account.ts
```diff
@@ -377,8 +381,21 @@ export class AzureLoginHelper {
 			const config = workspace.getConfiguration('azure');
 			if (config.get('cloud') !== selected.environment.name) {
 				this.doLogin = true;
```
v-tsun, line 383: Move this.doLogin to the top as before, think this should be good to resolve Rachel's comment.

src/azure-account.ts
```diff
@@ -783,81 +803,68 @@ async function getEnvironments(): Promise<AzureAccountEnvironment[]> {
...
+					galleryEndpointUrl: endpoints.galleryEndpoint,
+					storageEndpointSuffix: armUrl.substring(armUrl.indexOf('.')),
+					keyVaultDnsSuffix: '.vault'.concat(armUrl.substring(armUrl.indexOf('.'))),
+					managementEndpointUrl: endpoints.authentication.audiences[0],
+					validateAuthority: getValidateAuthority(endpoints.authentication.loginEndpoint)
+				};
 			}
-		]
+		} catch {
+			new Error(localize("azure-account.armUrlFetchFailed", "Fetch Azure custom cloud endpoints metadata failed"));
```
v-tsun, line 849: Add some notification to let user know if the fetch is failed.

RMacfarlane: To show this to users, you can use `vscode.window.showErrorMessage`. It's also possible to have buttons on the notification, it would be good to allow navigating direction to the setting. Like this:

```
			const openSettings = localize('openSettings', "Open Settings");
			// Do not block on the result of the notification
			window.showErrorMessage(
				localize('azure-account.armUrlFetchFailed', "Fetching custom cloud environment data failed. Please check your custom cloud settings."),
				openSettings
			).then(result => {
				if (result === openSettings) {
					commands.executeCommand('workbench.action.openSettings', '@ext:ms-vscode.azure-account customCloud');
				}
			});
```

RMacfarlane (changes requested):

src/azure-account.ts
```diff
@@ -376,12 +380,23 @@ export class AzureLoginHelper {
...
+					}
+				}
+
 				// if outside of normal range, set ppe setting
-				config.update('cloud', selected.environment.name, getCurrentTarget(config.inspect('cloud')));
-			} else {
-				return this.login('loginToCloud');
+				await config.update('cloud', selected.environment.name, getCurrentTarget(config.inspect('cloud')));
 			}
+			return this.login('loginToCloud');
```
v-tsun, line 399: Hey @RMacfarlane , I did some investigation, I think this doLogin is used to let us call logIn function inside initial function when we aren't logged. If we set multiple this.doLogIn = true, we will call logIn function multiple times, and will also trigger serval login pages. And I don't think it is reasonable if we call logIn function when we ain't finish setting the user setting. So I just change this piece of code a little bit just call logIn function in the end after we finish our config updates. What do you think about it ?

src/azure-account.ts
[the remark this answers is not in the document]
v-tsun: I add a new comment about this, please check it out

src/azure-account.ts
RMacfarlane: Thanks, this is a good solution

RMacfarlane (approved):
Looks good to me! Thanks!

PR #258, opened 2021-03-19
Add tenant intput box to improve user experience

v-tsun:
Hey @RMacfarlane, this PR add another inputbox for tenant Id when users are trying to login to Azure Custom Cloud without setting parameters in usersetting.json. Then users don't need to set tenant Id separately in user setting, we believe it would improve user's using experience.

### DIFF
--- a/src/azure-account.ts
+++ b/src/azure-account.ts
@@ -380,8 +380,6 @@ export class AzureLoginHelper {
 		if (selected) {
 			const config = workspace.getConfiguration('azure');
 			if (config.get('cloud') !== selected.environment.name) {
-				let tenantInputText = commonTenantId;
-				let tenantId;
 				let armUrl;
 				if (selected.environment.name === azureCustomCloud) {
 					armUrl = await window.showInputBox({
@@ -389,15 +387,14 @@ export class AzureLoginHelper {
 						placeHolder: 'https://management.local.azurestack.external',
 						ignoreFocusOut: true
 					});
-					tenantInputText = "Azure Custom Cloud Tenant Id";
 					if (!armUrl) {
 						// directly return when user didn't type in anything or press esc for resourceManagerEndpointUrl inputbox
 						return;
 					}
 				}
-				tenantId = await window.showInputBox({
+				const tenantId = await window.showInputBox({
 					prompt: localize('azure-account.enterTenantId', "Enter the tenant id"),
-					placeHolder: tenantInputText,
+					placeHolder: localize('azure-account.tenantIdPlaceholder', "Enter your tenant id, or '{0}' for the default tenant", commonTenantId),
 					ignoreFocusOut: true});
 				if (tenantId) {
 					if (armUrl) {
--- a/src/azure-account.ts
+++ b/src/azure-account.ts
@@ -397,8 +397,11 @@ export class AzureLoginHelper {
 					} else {
 						return;
 					}
+				} else {
+					// Set tenant Id back to commonTenantId if user isn't using Azure Custom Cloud to avoid unmatched tenant id issue.
+					await config.update('tenant', commonTenantId, getCurrentTarget(config.inspect('tenant')));
 				}
-
+				
 				// if outside of normal range, set ppe setting
 				await config.update('cloud', selected.environment.name, getCurrentTarget(config.inspect('cloud')));
 			}
--- a/src/azure-account.ts
+++ b/src/azure-account.ts
@@ -388,8 +388,8 @@ export class AzureLoginHelper {
 					});
 					if (armUrl) {
 						const tenantId = await window.showInputBox({
-							prompt: localize('azure-account.enterTenantId', "Enter the Tenant Id"),
-							placeHolder: "Tenant Id..",
+							prompt: localize('azure-account.enterTenantId', "Enter the tenant id"),
+							placeHolder: "Tenant Id",
 							ignoreFocusOut: true
 						});
 						if (tenantId) {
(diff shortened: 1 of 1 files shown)


src/azure-account.ts
```diff
@@ -386,8 +386,14 @@ export class AzureLoginHelper {
 						placeHolder: 'https://management.local.azurestack.external',
 						ignoreFocusOut: true
 					});
+					const tenantId = await window.showInputBox({
```
bganapa, line 389: where will this input box be shown?, if we do it for TenantId, why not for ARm endpoint as well?

src/azure-account.ts
```diff
@@ -386,8 +386,14 @@ export class AzureLoginHelper {
...
+						prompt: localize('azure-account.enterTenantId', "Enter the Tenant Id"),
+						placeHolder: "Please enter your tenant Id here",
+						ignoreFocusOut: true
+					});
 					if (armUrl) {
 						await config.update(customCloudArmUrlKey, armUrl, getCurrentTarget(config.inspect(customCloudArmUrlKey)));
+						await config.update('tenant', tenantId, getCurrentTarget(config.inspect('tenant')));
```
bganapa, line 396: I dont see the TenantId getting persisted to user settings

v-tsun:
hey @RMacfarlane maybe you can please take a look on this PR, we add a new inputbox for tenant id to improve user experience when using Azure Custom Cloud for first time. User can directly set properties they need through those inputboxes.

src/azure-account.ts
```diff
@@ -387,10 +387,23 @@ export class AzureLoginHelper {
 						ignoreFocusOut: true
 					});
 					if (armUrl) {
-						await config.update(customCloudArmUrlKey, armUrl, getCurrentTarget(config.inspect(customCloudArmUrlKey)));
+						const tenantId = await window.showInputBox({
+							prompt: localize('azure-account.enterTenantId', "Enter the Tenant Id"),
```
RMacfarlane, line 391: I think "Tenant Id" should be lowercase

src/azure-account.ts
```diff
@@ -387,10 +387,23 @@ export class AzureLoginHelper {
...
+							placeHolder: "Tenant Id..",
```
RMacfarlane, line 392: "Tenant id", extra period here

src/azure-account.ts
```diff
@@ -387,10 +387,23 @@ export class AzureLoginHelper {
...
+							ignoreFocusOut: true
+						});
+						if (tenantId) {
```
RMacfarlane, line 395: Will the user always have a specific tenant id when configuring a custom cloud? Or would there be times when using "common" is ok? If "common" is an option, maybe the placeholder text should give it as an example

src/azure-account.ts
```diff
@@ -387,10 +387,23 @@ export class AzureLoginHelper {
...
+							await config.update('tenant', tenantId, getCurrentTarget(config.inspect('tenant')));
+						} else {
+							return;
+						}
 					} else {
 						return;
 					}
+				} else {
+					// Set tenant Id back to commonTenantId if user isn't using Azure Custom Cloud to avoid unmatched tenant id issue.
+					await config.update('tenant', commonTenantId, getCurrentTarget(config.inspect('tenant')));
```
RMacfarlane, line 406: I'm wary of reseting this - I don't think we want to clear this. A user could have set a tenant setting and be using a different cloud

src/azure-account.ts
RMacfarlane: this should also be localized

RMacfarlane (changes requested):

src/azure-account.ts
v-tsun: Do we need to localized the placeholder?

src/azure-account.ts
v-tsun: Yeah, as far as my concern, I believe each Custom Cloud will have each specific tenant id, and I don't think common is an option we can use for Custom Cloud. @bganapa Please correct me if I am wrong.

src/azure-account.ts
v-tsun: For this piece of code, you can see from its context https://github.com/microsoft/vscode-azure-account/blob/917fdc42cd761e63ea045a8d38d7a5cbac369cd9/src/azure-account.ts#L382-L407
It will only be called when the cloud user selects is different from the one we have in config and that cloud can not be Azure Custom Cloud. 
A user could set a tenant id can be used for serval cloud. But sometime the tenant id won't be same, for example, the tenant id of Azure Custom Cloud may not work for the other clouds. For example, if we try to sign in to Azure if we have Azure Custom Cloud tenant id in our setting. Error as below will be shown. 
image
That is why I reset tenant id here. 
Will clouds like Azure, AzureChina... have specific tenant id other than common? If so, I think we probably could make the tenant inputbox available every time user selects a cloud which is different from the one we are having in config, not only when user is selecting Azure Custom Cloud. What do you think, Rachel? 
@bganapa, could you also take a look and give us some feedback if you have some idea about this.

src/azure-account.ts
v-tsun: And also Azure Custom Cloud would have different tenant id for AAD and ADFS, therefore, it isn't that easy to just give a simple placeholder to represent both.

bganapa: Yes, common is never an option for azure stack, it will always have a specific tenant id in case of AAD. in ADFS case, its just 'adfs'

src/azure-account.ts
RMacfarlane: Yeah, for other clouds users can potentially configure their own tenant id. I think it would make sense to show the tenant input box every time the user chooses a different cloud so they can change it to the correct value. I think the majority of the time for the other clouds they would be using `common`, though, so it would be nice if we prefilled the input box with `common` then and left it empty for Azure Custom Cloud.

RMacfarlane (approved):
Made a slight tweak to the placeholder text for the tenant input, this looks good!

bganapa:
Thanks @RMacfarlane !

PR #272, opened 2021-06-01
npm update, fix tsc errors

wwlorey:


src/codeFlowLogin.ts
```diff
@@ -231,7 +231,7 @@ function createTerminateServer(server: http.Server) {
 		});
 	});
 	return async () => {
-		const result = new Promise<void>(resolve => server.close(resolve));
+		const result = new Promise<void>(() => server.close());
```
ejizba, line 234: What's the reasoning behind this change? Not passing the `resolve` could change the timing, possibly causing a race condition

src/codeFlowLogin.ts
```diff
@@ -231,7 +231,7 @@ function createTerminateServer(server: http.Server) {
...
 		for (const id in sockets) {
```
ejizba, line 235: Do we have test cases for this extension and if so where are they (I didn't see them in the OneNote)? With all these engineering changes lately, we need to make sure we have some regular testing to reduce the chance of regressions

src/codeFlowLogin.ts
```diff
@@ -340,27 +341,25 @@ async function startServer(server: http.Server, adfs: boolean) {
 	return port;
 }
 
-/* eslint-disable @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access */
 async function callback(nonce: string, reqUrl: url.Url): Promise<string> {
-	let error = reqUrl.query.error_description || reqUrl.query.error;
```
ejizba, line 345: How come we're no longer checking the query for an error?

src/codeFlowLogin.ts
```diff
@@ -231,7 +232,10 @@ function createTerminateServer(server: http.Server) {
 		});
 	});
 	return async () => {
-		const result = new Promise<void>(resolve => server.close(resolve));
+		const result = new Promise<void>((resolve, reject) => {
+			server.close(reject)
+			resolve();
```
ejizba, line 237: By calling `resolve` here, that means the `new Promise` will resolve immediately instead of waiting until the `server.close` event happens, which could again cause timing issues.

This is what we normally do with an `err?: Error | undefined` callback:
```typescript
const result = new Promise<void>((resolve, reject) => server.close(error => {
	if (error) {
		reject(error);
	} else {
		resolve();
	}
}));
```
_However_, that piece of code would throw an error whereas the original code would ignore the error. It's entirely possible ignoring the error was intentional behavior and that's why I do _not_ want to do this during an engineering PR. Most of this PR is sorting imports and messing with the types, which can't affect functionality. We can mess with the types to fix this line, too:

```typescript
const result = new Promise<void>((resolve: () => void) => server.close(resolve));
```

src/codeFlowLogin.ts
```diff
@@ -340,27 +342,32 @@ async function startServer(server: http.Server, adfs: boolean) {
...
-		const receivedNonce = (state.split(',')[1] || '').replace(/ /g, '+');
-		if (receivedNonce !== nonce) {
-			error = 'Nonce does not match.';
+	let query: ParsedUrlQuery;
+	let error: string | undefined;
+	let code: string | undefined;
+
+	if (reqUrl.query) {
+		query = typeof reqUrl.query === 'string' ? parse(reqUrl.query) : reqUrl.query;
+		error = query.error_description.toString() || query.error.toString() || '';
```
ejizba, line 352: I believe the properties on `query` will be `undefined` if they're not specified in the url, so by adding `toString()` you could be causing a null ref error. There's a lot of ways to fix that, but I'd probably end up writing a little helper function
```typescript
function getQueryProp(query: ParsedUrlQuery, propName: string): string {
    const value = query[propName];
    return typeof value === 'string' ? value : '';
}
```
```typescript
error = getQueryProp(query, 'error_description') || getQueryProp(query, 'error');
```

ejizba (approved):

src/codeFlowLogin.ts
ejizba: Please make sure the tests cover both methods of signing in ("sign in" and "sign in with device code"). And make sure that CTI is running these as a part of their regular testing

wwlorey: Will do 👍

PR #278, opened 2021-06-11
Split up azure-account.ts into multiple files

wwlorey:
This is the first of several PRs to modularize the code before we introduce any MSAL stuff. I didn't change any code I moved, the only things that changed were some ESLint exceptions here and there.

I put many of the new files directly in `src` for now but I'll move them to a subfolder when I rename `azure-account.ts`. And I made sure the manual tests are passing for these changes

### DIFF
--- a/src/azure-account.ts
+++ b/src/azure-account.ts
@@ -13,13 +13,13 @@ import { AzureAccount, AzureLoginStatus, AzureResourceFilter, AzureSession, Azur
 import { becomeOnline } from './checkIsOnline';
 import { createCloudConsole } from './cloudConsole';
 import * as codeFlowLogin from './codeFlowLogin';
-import { azureCustomCloud, azurePPE, clientId, commonTenantId, customCloudArmUrlKey, environmentLabels, staticEnvironmentNames, staticEnvironments } from './constants';
+import { azureCustomCloud, azurePPE, clientId, commonTenantId, customCloudArmUrlKey, staticEnvironments } from './constants';
 import { deviceLogin } from './deviceLogin';
 import { getEnvironments, getSelectedEnvironment } from './environments';
 import { AzureLoginError, getErrorMessage } from './errors';
 import { addFilter, getNewFilters, removeFilter } from './filters';
 import { TelemetryReporter } from './telemetry';
-import { addTokenToCache, Cache, clearTokenCache, deleteRefreshToken, getStoredCredentials, ProxyTokenCache, storeRefreshToken, tokenFromRefreshToken, tokensFromToken } from './tokens';
+import { addTokenToCache, clearTokenCache, deleteRefreshToken, getStoredCredentials, ProxyTokenCache, storeRefreshToken, tokenFromRefreshToken, tokensFromToken } from './tokens';
 import { listAll } from './utils/arrayUtils';
 import { localize } from './utils/localize';
 import { openUri } from './utils/openUri';
--- a/src/azure-account.ts
+++ b/src/azure-account.ts
@@ -13,7 +13,7 @@ import { AzureAccount, AzureLoginStatus, AzureResourceFilter, AzureSession, Azur
 import { becomeOnline } from './checkIsOnline';
 import { createCloudConsole } from './cloudConsole';
 import * as codeFlowLogin from './codeFlowLogin';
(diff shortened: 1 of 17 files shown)


src/azure-account.ts
```diff
@@ -3,213 +3,40 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-// eslint-disable-next-line @typescript-eslint/no-var-requires, @typescript-eslint/no-unsafe-assignment
```
ejizba, line 6: How does the resulting vsix compare after your changes? Specifically, I want to make sure the result of a webpack looks about the same (same number of files, same approximate size of the files) so that we're not affecting performance

Also when you say "I made sure the manual tests are passing for these changes" were you F5'ing or running against the webpacked vsix? Please make sure to at least spot check with the webpacked vsix

src/azure-account.ts
```diff
@@ -3,213 +3,40 @@
...
-const CacheDriver = require('adal-node/lib/cache-driver');
-// eslint-disable-next-line @typescript-eslint/no-var-requires, @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access
-const createLogContext = require('adal-node/lib/log').createLogContext;
-
-import { SubscriptionClient, SubscriptionModels } from '@azure/arm-subscriptions';
```
ejizba, line 11: Can we get a quick "outline" of what you want the resulting file structure to look like after you've moved all files and after you've added msal code? None of these names seem bad, but it's hard to get a big picture idea of what the end goal is.

I don't really care how you create the outline, but personally I would probably mess around with some empty files/folders in VS Code to get a screenshot like this:
Screen Shot 2021-06-11 at 10 48 09 AM

src/azure-account.ts
wwlorey: This PR's vsix is 755k compared to 754k before the changes. And the old & new vsixes have the same number of files.

I was F5ing but I just verified the test cases with the vsix as well. I had to include a new dev dependency that vsce complained about.

src/azure-account.ts
wwlorey: Here's a rough look at what I'm thinking: 

Screen Shot 2021-06-11 at 11 28 20 AM

src/azure-account.ts
ejizba: I don't think renaming files would related to dev dependencies at all. Seems fishy

wwlorey: This was the first time I tried running `vsce package`. I get the same error on main

src/constants.ts
```diff
@@ -0,0 +1,36 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { Environment } from "@azure/ms-rest-azure-env";
+import { localize } from "./utils/localize";
+
```
ejizba, line 8: Just make sure we're not throwing every single constant in here. IMO this file should only be used for constants shared across multiple files or "noteworthy" constants

src/tokens.ts
```diff
@@ -0,0 +1,189 @@
...
+import { localize } from "./utils/localize";
+
+// eslint-disable-next-line @typescript-eslint/no-var-requires, @typescript-eslint/no-unsafe-assignment
+const CacheDriver = require('adal-node/lib/cache-driver');
+// eslint-disable-next-line @typescript-eslint/no-var-requires, @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access
+const createLogContext = require('adal-node/lib/log').createLogContext;
+
+const keytar = tryGetKeyTar();
+
+export interface Cache {
```
ejizba, line 23: I don't think this interface is related to tokens

src/checkIsOnline.ts
```diff
@@ -0,0 +1,39 @@
+import { Environment } from "@azure/ms-rest-azure-env";
```
ejizba, line 1: copyright header

ejizba (approved):

PR #279, opened 2021-06-14
Engineering changes for login-related files and utils

wwlorey:
I'll rename these files as needed in the next PR


ejizba (approved):
Would've preferred a bit more descriptive title and/or description. If I understand the PR correctly, maybe "Improve names and add types" would be accurate. Just something to give me an idea how likely it is that the engineering changes could affect functionality

src/azure-account.ts
```diff
@@ -40,19 +40,17 @@ const environmentLabels: Record<string, string> = {
 	[azurePPE]: localize('azure-account.azurePPE', 'Azure PPE'),
 };
 
-const logVerbose = false;
+const logVerbose: boolean = false;
 
-interface AzureAccountWriteable extends AzureAccount {
+interface IAzureAccountWriteable extends AzureAccount {
```
neelip, line 45: Should we follow this rule of prefixing interfaces with an "I" given that eslint doesn't require this anymore? 
If we plan to, we should probably add this rule for eslint checks. 
https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/naming-convention.md#enforce-that-interface-names-do-not-begin-with-an-i

PR #282, opened 2021-06-14
Add constants and helper functions for settings

wwlorey:


src/constants.ts
```diff
@@ -5,12 +5,19 @@
 
 import { Environment } from "@azure/ms-rest-azure-env";
 
+export const prefix: string = 'azure';
```
ejizba, line 8: nit: we usually call this `extensionPrefix`

ejizba (approved):

PR #285, opened 2021-06-15
Add types and update identifier names in AzureLoginHelper.ts

wwlorey:


src/AzureLoginHelper.ts
```diff
@@ -358,31 +364,31 @@ export class AzureLoginHelper {
 				subscription
 			}))
 		}
-		void this.context.globalState.update('cache', cache);
+		void this.context.globalState.update(cacheKey, cache);
 	}
 
-	private beginLoggingIn() {
+	private setLoggedInStatus(): void {
```
ejizba, line 370: I liked `beginLoggingIn` better. Currently, `setLoggedInStatus` and `updateLoginStatus` are too similar

src/AzureLoginHelper.ts
```diff
@@ -506,40 +512,39 @@ export class AzureLoginHelper {
...
 		return subscriptions;
 	}
 
-	private asSubscriptionItems(subscriptions: AzureSubscription[], resourceFilter: string[]): ISubscriptionItem[] {
+	private getSubscriptionItems(subscriptions: AzureSubscription[], resourceFilter: string[]): ISubscriptionItem[] {
 		return subscriptions.map(subscription => {
-			const picked = resourceFilter.indexOf(`${subscription.session.tenantId}/${subscription.subscription.subscriptionId}`) !== -1 || resourceFilter[0] === 'all';
+			const picked: boolean = resourceFilter.indexOf(`${subscription.session.tenantId}/${subscription.subscription.subscriptionId}`) !== -1 || resourceFilter[0] === 'all';
 			return <ISubscriptionItem>{
-				type: 'item',
```
ejizba, line 542: Any idea why `type: 'item'` was on here in the first place? Getting rid of it may affect functionality, and I don't really trust the types in this repo to verify if/where it's being used

ejizba: I couldn't find it used anywhere so I'll go ahead and approve, but wouldn't mind you double-checking as well

ejizba (approved):

src/AzureLoginHelper.ts
wwlorey: I looked as well and couldn't find any reference to it. I thought `type` might be an artifact from an old version of the `QuickPickItem` interface

PR #287, opened 2021-06-16
Update identifier names and add types for login.ts and related code

wwlorey:


src/server.ts
```diff
@@ -12,24 +12,30 @@ import { parse, ParsedUrlQuery } from 'querystring';
 import * as url from 'url';
 import { portADFS, redirectUrlAAD } from './constants';
 
-interface Deferred<T> {
+interface IDeferred<T> {
```
ejizba, line 15: Ftr, prefixing interfaces with "I" is another rule that was enabled by default in tslint, but not in eslint. You don't have to use an "I" if you don't want to and I personally plan to stop doing it. No one else really uses "I" (VS Code types, Node.js types, etc.) and it bothers me that "interface" has an "I", but "type" doesn't even though they're very similar.

ejizba (approved):

PR #297, opened 2021-07-27
Add auth library abstraction

wwlorey:
These changes don't introduce any MSAL code, I'm just moving the existing ADAL code into an abstraction layer. I'll move login.ts, tokens.ts, and loginWithDeviceCode.ts into the ADAL folder in a separate PR (those files are only used with ADAL).

### DIFF
--- a/src/login/AzureLoginHelper.ts
+++ b/src/login/AzureLoginHelper.ts
@@ -5,33 +5,25 @@
 
 import { SubscriptionClient, SubscriptionModels } from '@azure/arm-subscriptions';
 import { Environment } from '@azure/ms-rest-azure-env';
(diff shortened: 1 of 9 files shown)


src/login/AzureLoginHelper.ts
```diff
@@ -97,6 +90,8 @@ export class AzureLoginHelper {
 	};
 
 	constructor(private context: ExtensionContext, private reporter: TelemetryReporter) {
+		this.authProvider = new AbstractAuthProvider(context, enableVerboseLogs);
```
ejizba, line 93: The constructor of an abstract class is generally not accessible. You would do `new AdalAuthProvider()` or `new MsalAuthProvider()` and use those directly

ejizba (approved):

Issue #203, opened 2020-07-10
Investigate switching to MSAL JS library


bwateratmsft:
It looks like the `adal-node` dependency might be one of the last ones that depends on `request` (which is deprecated). The Azure SDK for JS replaces it with `node-fetch`, looks like MSAL with `axios`.

It may also fix #53 according to [this comment](https://github.com/microsoft/vscode-azure-account/issues/53#issuecomment-521584521).

fhljys:
Any update?

albertleigh:
Any plans? 😀

wwlorey:
Support for MSAL will be released in version 0.10.0 of the extension.

bwateratmsft:
Woohoo!

ruiminwang:
> 

When will 0.10.0 be released? Looking forward to the new version!

wwlorey:
Support for MSAL has been released! 🎉 In [v0.10.0](https://github.com/microsoft/vscode-azure-account/releases/tag/v0.10.0) you can now use the `Azure: Authentication Library` setting to switch between ADAL and MSAL.

bwateratmsft:
I'm going to reopen this as it's active again.

shsuman:
Are there any updates on this issue ?

PR #299, opened 2021-07-28
Add `MsalAuthProvider`

wwlorey:


src/login/adal/AdalAuthProvider.ts
```diff
@@ -40,10 +34,10 @@ export class AdalAuthProvider extends AuthProviderBase {
 			// eslint-disable-next-line @typescript-eslint/no-explicit-any
 			log: (_level: any, message: any, error: any) => {
 				if (message) {
-					super.outputChannel.appendLine(message);
+					this.outputChannel.appendLine(message);
```
wwlorey, line 37: Apparently `outputChannel` isn't accessible on `super` even though it compiles  🤷‍♂️

src/login/msal/MsalAuthProvider.ts
```diff
@@ -0,0 +1,122 @@
...
+					piiLoggingEnabled: false,
+					logLevel: enableVerboseLogs ? LogLevel.Verbose : LogLevel.Error
+				}
+			}
+		};
+		this.publicClientApp = new PublicClientApplication(msalConfiguration);
+	}
+
+	public async loginWithoutLocalServer(_clientId: string, _environment: Environment, _isAdfs: boolean, _tenantId: string): Promise<AbstractLoginResult> {
+		throw new Error('"Login Without Local Server" not implemented for MSAL.');
```
wwlorey, line 46: To avoid bloating this PR too much I'm waiting to implement `loginWithoutLocalServer` and `loginWithDeviceCode` for MSAL

src/login/msal/MsalAuthProvider.ts
```diff
@@ -0,0 +1,122 @@
...
+			accountInfo: loginResult.account!,
+			// eslint-disable-next-line @typescript-eslint/no-explicit-any
+			credentials: this.getCredentials(<any>environment, loginResult.account!.username, loginResult.tenantId),
+			credentials2: this.getCredentials2(environment, loginResult.account!.username, loginResult.tenantId, loginResult.account!)
+		});
+		/* eslint-enable @typescript-eslint/no-non-null-assertion */
+	}
+
+	public async clearLibraryTokenCache(): Promise<void> {
+		// MSAL handles token caching under the hood
```
ejizba, line 119: They _can_ handle stuff under the hood for you, but you have to let them. Specifically, I think you need to call `TokenCache.removeAccount`. I believe that would be more robust than the stuff you're currently doing in `clearLocalTokenCache`, which should likely only be used for ADAL (aka just revert most of your changes that moved `AdalAuthProvider.clearTokenCache` into `AuthProviderBase.clearLocalTokenCache`).

src/login/AzureLoginHelper.ts
```diff
@@ -320,7 +327,7 @@ export class AzureLoginHelper {
 	}
```
ejizba, line 328: I think we should remove `getStoredCredentials` from this file and put it into `AdalAuthProvider.loginSilent` since MSAL doesn't need it. In fact, you should be able to remove all references to `credentialsSection` and `tryGetKeyTar` in the "common" files (aka anything outside the `msal` and `adal` folders) since each library has their own caching mechanism (MSAL does it through `cachePlugin` and ADAL makes us do everything manually). My comment around `TokenCache.removeAccount` is related to this, btw.

src/azure-account.api.d.ts
```diff
@@ -40,7 +43,7 @@ export interface AzureSession {
 	/**
 	 * The credentials object for azure-sdk-for-js modules https://github.com/azure/azure-sdk-for-js
 	 */
-	readonly credentials2: TokenCredentialsBase;
+	readonly credentials2: TokenCredentialsBase | AzureIdentityCredentialAdapter;
```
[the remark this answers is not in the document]
wwlorey: There are some ADAL specific properties on [`TokenCredentialsBase`](https://github.com/Azure/ms-rest-nodeauth/blob/684d9825afe25fc907e8d163d19c0996702d0a70/lib/credentials/tokenCredentialsBase.ts#L16) that are not present on [`AzureIdentityCredentialAdapter`](https://github.com/Azure/ms-rest-js/blob/f6f0d92d79a1dfa8d92ff0891b88bc6b7a349e69/lib/credentials/azureIdentityTokenCredentialAdapter.ts#L17) so it could definitely break people. What do you think about creating a `credentials3` that's just type `AzureIdentityCredentialAdapter`? I don't think there's a way to fully avoid making a breaking change

src/azure-account.api.d.ts
```diff
@@ -31,6 +33,7 @@ export interface AzureSession {
 	readonly environment: Environment;
 	readonly userId: string;
 	readonly tenantId: string;
+	readonly accountInfo?: AccountInfo;
```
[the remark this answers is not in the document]
wwlorey: With ADAL we need the cached `environment`, `userId`, and `tenantId` to create credentials
https://github.com/microsoft/vscode-azure-account/blob/c47228517bb1670691b325c0ac9c79415abdf922/src/login/adal/AdalAuthProvider.ts#L127

But creating credentials with MSAL requires `accountInfo` so we have to cache it
 https://github.com/microsoft/vscode-azure-account/blob/c47228517bb1670691b325c0ac9c79415abdf922/src/login/msal/MsalAuthProvider.ts#L95

src/login/AuthProviderBase.ts
```diff
@@ -4,22 +4,42 @@
...
+const keytar: KeyTar | undefined = tryGetKeyTar();
+
+export type AbstractLoginResult = TokenResponse[] | AuthenticationResult;
 export type AbstractCredentials = DeviceTokenCredentials;
-export type AbstractCredentials2 = DeviceTokenCredentials2;
+export type AbstractCredentials2 = DeviceTokenCredentials2 | AzureIdentityCredentialAdapter;
+
+export const loginResultTypeError: Error = new Error(localize('azure-account.unexpectedType', 'Unexpected login result type.'));
+
+export function isAdalLoginResult(loginResult: AbstractLoginResult): boolean {
```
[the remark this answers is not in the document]
wwlorey: This is nifty. And I didn't have the setting turned on for whitespace formatting but I do now 👍

src/azure-account.api.d.ts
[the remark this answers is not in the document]
ejizba: My question is specific to the "api", though. Do external extensions need it - or is it only used within the azure account extension?

src/azure-account.api.d.ts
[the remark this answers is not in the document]
ejizba: If we have to do a breaking change regardless, I would love if we can switch over to a versioned api, like this:

src/azure-account.api.d.ts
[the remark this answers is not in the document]
wwlorey: Oh I see - no it's not needed externally. I'll remove it

src/azure-account.api.d.ts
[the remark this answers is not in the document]
ejizba: To be clear, though, I think switching to a versioned API is probably best done in it's own PR.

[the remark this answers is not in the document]
wwlorey: Tracking this here: https://github.com/microsoft/vscode-azure-account/issues/304

webpack.config.js
```diff
@@ -59,7 +60,12 @@ module.exports = (_, argv) => {
 							}
 						}
 					}]
-			}]
+				},
+				{
+					test: /\.node$/,
+					loader: "node-loader"
```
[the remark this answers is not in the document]
wwlorey: I thought this was required after adding some dependencies but turns out it's not - reverted this file

wwlorey (commented):
I found a bug while I was working on these changes: if you were signed in with ADAL then switched to MSAL without signing out (after reloading the window), your MSAL login wouldn't persist after a refresh.

So I added a configuration listener that signs you out if you change the auth library setting.

package.json
```diff
@@ -200,6 +200,19 @@
           },
           "default": null,
           "description": "Development setting: The PPE environment for testing."
+        },
+        "azure.authenticationLibrary": {
+          "type": "string",
+          "enum": [
+            "ADAL",
+            "MSAL (Preview)"
```
ejizba, line 208: I think "Preview" is more of a human-readable description than a programmatic value, so I would only include it in "enumDescriptions", not "enum". For example, eventually "MSAL" won't be preview anymore, but we still want a user's hard-coded settings to continue working

package.json
```diff
@@ -200,6 +200,19 @@
...
+          ],
+          "enumDescriptions": [
+            "Azure Active Directory Authentication Library",
+            "Microsoft Authentication Library (Preview)"
+          ],
+          "default": "ADAL",
```
ejizba, line 214: I would not specify a default in package.json since we will likely be changing it in the near future. Instead, I would handle the default logic inside our typescript code - which is actually where any A/B framework stuff would go if we end up doing that.

src/login/AuthProviderBase.ts
```diff
@@ -4,24 +4,28 @@
...
 import { getKey } from "./getKey";
 import { CodeResult, createServer, createTerminateServer, RedirectResult, startServer } from './server';
 
-export type AbstractLoginResult = TokenResponse[];
 export type AbstractCredentials = DeviceTokenCredentials;
-export type AbstractCredentials2 = DeviceTokenCredentials2;
+export type AbstractCredentials2 = DeviceTokenCredentials2 | AzureIdentityCredentialAdapter;
 
-export abstract class AuthProviderBase {
+export const loginResultTypeError: Error = new Error(localize('azure-account.unexpectedType', 'Unexpected login result type.'));
```
ejizba, line 26: Remove this - it's no longer used.

src/login/AuthProviderBase.ts
```diff
@@ -31,32 +35,32 @@ export abstract class AuthProviderBase {
...
+	public abstract loginWithDeviceCode(environment: Environment, tenantId: string): Promise<TLoginResult>;
+	public abstract loginSilent(environment: Environment, tenantId: string, migrateToken?: boolean): Promise<TLoginResult>;
 	public abstract getCredentials(environment: string, userId: string, tenantId: string): AbstractCredentials;
-	public abstract getCredentials2(environment: Environment, userId: string, tenantId: string): AbstractCredentials2;
-	public abstract updateSessions(environment: Environment, loginResult: AbstractLoginResult, sessions: AzureSession[]): Promise<void>;
-	public abstract clearTokenCache(): Promise<void>;
-	public abstract deleteRefreshTokens(): Promise<void>;
+	public abstract getCredentials2(environment: Environment, userId: string, tenantId: string, accountInfo?: AccountInfo): AbstractCredentials2;
+	public abstract updateSessions(environment: Environment, loginResult: TLoginResult, sessions: AzureSession[]): Promise<void>;
+	public abstract clearLibraryTokenCache(): Promise<void>;
```
ejizba, line 45: I don't think we need both `clearLibraryTokenCache` and `clearLocalTokenCache`. Just keep it all in one place (`clearTokenCache`) like it was before

src/login/AzureLoginHelper.ts
```diff
@@ -100,14 +105,24 @@ export class AzureLoginHelper {
...
+		context.subscriptions.push(workspace.onDidChangeConfiguration(async e => {
 			if (e.affectsConfiguration(getSettingWithPrefix(cloudSetting)) || e.affectsConfiguration(getSettingWithPrefix(tenantSetting)) || e.affectsConfiguration(getSettingWithPrefix(customCloudArmUrlSetting))) {
 				const doLogin: boolean = this.doLogin;
 				this.doLogin = false;
 				this.initialize(e.affectsConfiguration(getSettingWithPrefix(cloudSetting)) ? 'cloudChange' : e.affectsConfiguration(getSettingWithPrefix(tenantSetting)) ? 'tenantChange' : 'customCloudARMUrlChange', doLogin)
 					.catch(console.error);
 			} else if (e.affectsConfiguration(getSettingWithPrefix(resourceFilterSetting))) {
 				this.updateFilters(true);
+			} else if (e.affectsConfiguration(getSettingWithPrefix(authLibrarySetting))) {
+				await this.logout();
```
ejizba, line 117: It's a bit jarring to me that it immediately logs out as soon as the setting is changed. Maybe we can make it a bit smoother? Like change the message to "You must sign out and reload the window..."? and only sign out if they click the button

The other option is to support both ADAL and MSAL within the same session. You would need to create both auth providers in the constructor, and then here you would just change which one is used. I feel like that would either "just work" or it would get super complicated and not be worth it.

src/login/adal/AdalAuthProvider.ts
```diff
@@ -14,17 +14,20 @@ import { azureCustomCloud, azurePPE, clientId, redirectUrlAAD, staticEnvironment
...
-import { addTokenToCache, clearTokenCache, deleteRefreshToken, getTokenResponse, getTokensFromToken, getTokenWithAuthorizationCode, ProxyTokenCache, storeRefreshToken, tokenFromRefreshToken } from "./tokens";
+import { addTokenToCache, clearTokenCache, deleteRefreshToken, getStoredCredentials, getTokenResponse, getTokensFromToken, getTokenWithAuthorizationCode, ProxyTokenCache, storeRefreshToken, tokenFromRefreshToken } from "./tokens";
 
 const staticEnvironmentNames: string[] = [
 	...staticEnvironments.map(environment => environment.name),
 	azureCustomCloud,
-	azurePPE
+	azurePPE,
+	// 'Azure' and 'AzureChina' are the old names for the 'AzureCloud' and 'AzureChinaCloud' environments
+	'Azure',
```
ejizba, line 26: I don't like having the old names in a file-level const like this. We only want to use the old names in specific circumstances and I could see someone accidentally using this const for other reasons.

src/login/AzureLoginHelper.ts
wwlorey: I went with option 1. Option 2 gets dicey because we'd still have to reload the window if they change the setting - otherwise they'd still be signed in using the old auth library when interacting with the extensions

src/login/AzureLoginHelper.ts
```diff
@@ -527,26 +536,6 @@ function getCurrentTarget(config: { key: string; defaultValue?: unknown; globalV
...
-	} catch {
-		// ignore
-	}
-	try {
-		return await keytar.getPassword(credentialsSection, environment.name) || undefined;
-	} catch {
-		// ignore
-	}
+function getAuthLibrary(): AuthLibrary {
+	return getSettingValue<AuthLibrary>(authLibrarySetting) || 'MSAL';
```
ejizba, line 540: This should default to ADAL for now

src/login/msal/cachePlugin.ts
```diff
@@ -0,0 +1,37 @@
...
+
+const keytar: KeyTar | undefined = tryGetKeyTar();
+
+const beforeCacheAccess = async (cacheContext: TokenCacheContext): Promise<void> => {
+	if (keytar) {
+		try {
+            const cachedValue: string | null = await keytar.getPassword(credentialsSection, (await getSelectedEnvironment()).name);
+			cachedValue && cacheContext.tokenCache.deserialize(cachedValue);
+		} catch (error) {
+			console.log(JSON.stringify(error));
```
ejizba, line 19: I'm not a big fan of `console.log` in production code - it should either be in the output channel or not logged at all. Was this just here for debugging purposes? In any case, I don't think we want to log any errors for keytar/password-related stuff

wwlorey: I think that was an artifact of the sample code. Just removed it

ejizba (approved):

wwlorey:

Issue #304, opened 2021-08-18
Switch to versioned API


wwlorey:
From https://github.com/microsoft/vscode-azure-account/pull/299#discussion_r690723227

PR #305, opened 2021-08-19
Add versioned API

wwlorey:
Fixes https://github.com/microsoft/vscode-azure-account/issues/304


src/extension.ts
```diff
@@ -30,7 +32,8 @@ export async function activate(context: ExtensionContext): Promise<AzureAccount>
 	context.subscriptions.push(commands.registerCommand('azure-account.openCloudConsoleWindows', () => cloudConsole(azureLoginHelper.api, 'Windows')));
 	context.subscriptions.push(commands.registerCommand('azure-account.uploadFileCloudConsole', uri => uploadFile(azureLoginHelper.api, uri)));
 	survey(context, reporter);
-	return Promise.resolve(azureLoginHelper.api); // Return promise to work around weird error in WinJS.
```
wwlorey, line 33: This workaround was added 4 years ago in 6112d6dd75e6de51e568de69cc2ff12e0c8184bc so I feel comfortable removing it

package.json
```diff
@@ -259,6 +259,7 @@
     "request-promise": "4.2.2",
     "semver": "5.6.0",
     "uuid": "^8.3.2",
+    "vscode-azureextensionui": "^0.48.0",
```
ejizba, line 262: So far we've avoided using this package to keep the account extension small - especially as it relates to activation times. Can you look into the effects of using this package before we start using it for everything?

src/extension.ts
```diff
@@ -30,7 +32,8 @@ export async function activate(context: ExtensionContext): Promise<AzureAccount>
...
+
+	return createApiProvider([azureLoginHelper.api]);
```
ejizba, line 36: We need to have a transition period between the versioned api and the old api. One way to accomplish that would be something like `return Object.assign(createApiProvider([azureLoginHelper.api]), azureLoginHelper.api);`. However that's a bit simplistic because ideally any _new_ stuff would only go in the versioned api (somewhat as an incentive for people to move to the versioned api)

Also would be nice to have the following:
- A deprecation warning to extension authors using the old api (maybe in the console? idk)
- Telemetry on how many people are using the old api vs the versioned api

package.json
wwlorey: I did some activation time testing on my machine and adding the UI package doesn't seem to make an appreciable difference (units are ms):

| VSIX with UI package | VSIX without UI package |
|------------|---------------|
| 98         | 65            |
| 65         | 62            |
| 63         | 68            |
| 64         | 68            |
| 65         | 66            |

ejizba: Can you check on Windows as well? If that checks out - I'm okay using the UI package. However we need to keep on eye on this, specifically as we use more and more of the package and as we get telemetry on activation times from real users (we have that telemetry in this extension, right?)

src/login/AzureLoginHelper.ts
```diff
@@ -479,39 +511,61 @@ export class AzureLoginHelper {
...
+		return this.useLegacyApi ? this.legacyApi : this.api;
+	}
+
+	private setUseLegacyApi(eventName: string, useLegacyApi?: boolean) {
+		const useLegacyApiOld: boolean | undefined = this.useLegacyApi;
+		this.useLegacyApi = this.useLegacyApi === undefined ? useLegacyApi : this.useLegacyApi;
+
+		if (this.useLegacyApi && this.useLegacyApi !== useLegacyApiOld) {
+			// Only log a message once when `useLegacyApi` is set for the first time
+			ext.outputChannel.appendLog(localize('azure-account.depreciationWarning', 'WARNING: You are using a depreciated version of the Azure Account API. Upgrade for access to the latest features.'));
```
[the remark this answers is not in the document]
wwlorey: I went ahead and removed this log. I agree filing an issue on the dependent extension's repos is more effective

src/login/AzureAccountInternal.ts
```diff
@@ -0,0 +1,162 @@
...
+import { AzureAccount, AzureLoginStatus, AzureResourceFilter, AzureSession, AzureSubscription, CloudShell } from '../azure-account.api';
+import { extensionPrefix, resourceFilterSetting } from '../constants';
+import { TelemetryReporter } from '../telemetry';
+import { localize } from '../utils/localize';
+import { openUri } from '../utils/openUri';
+import { AzureLoginHelperTasks, AzureLoginStatusObj, ISubscriptionItem } from './AzureLoginHelper';
+import { addFilter, removeFilter } from './filters';
+import { getCurrentTarget } from './getCurrentTarget';
+
+export class AzureAccountInternal implements AzureAccount {
```
ejizba, line 16: Per offline discussion, use a separate legacy api class that references this class

src/login/AzureAccountInternal.ts
```diff
@@ -0,0 +1,162 @@
...
+import { CancellationTokenSource, commands, ConfigurationTarget, Disposable, Event, MessageItem, QuickPickItem, window, workspace, WorkspaceConfiguration } from 'vscode';
+import { AzureAccount, AzureLoginStatus, AzureResourceFilter, AzureSession, AzureSubscription, CloudShell } from '../azure-account.api';
+import { extensionPrefix, resourceFilterSetting } from '../constants';
+import { TelemetryReporter } from '../telemetry';
+import { localize } from '../utils/localize';
+import { openUri } from '../utils/openUri';
+import { AzureLoginHelperTasks, AzureLoginStatusObj, ISubscriptionItem } from './AzureLoginHelper';
+import { addFilter, removeFilter } from './filters';
+import { getCurrentTarget } from './getCurrentTarget';
+
```
ejizba, line 15: My preferred naming would be:
```
import * as types from '../azure-account.api';

export class AzureAccountExtensionApi implements types.AzureAccountExtensionApi {
```

and

```
import * as legacyTypes from '../azure-account.legacy.api';

export class AzureAccountExtensionLegacyApi implements legacyTypes.AzureAccount {
```

src/login/msal/MsalAuthProvider.ts
```diff
@@ -87,9 +81,8 @@ export class MsalAuthProvider extends AuthProviderBase<AuthenticationResult> {
 		}
 	}
 
-	public getCredentials(environment: string, userId: string, tenantId: string): DeviceTokenCredentials {
-		// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-explicit-any
-		return new DeviceTokenCredentials({ environment: (<any>Environment)[environment], username: userId, clientId, tokenCache: this.dummyTokenCache, domain: tenantId });
+	public getCredentials(): undefined {
+		return undefined;
```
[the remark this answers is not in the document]
wwlorey: I changed this back to `return undefined` because throwing an error would prevent the cached MSAL session from being loaded in `initializeSessions`. Do you know if there's a way to only throw an error only if that credentials object is accessed?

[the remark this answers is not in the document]
ejizba: You'd want to use a getter. You can do than on any object, but the cleanest option would likely be to create an `AzureSession` class

src/login/AzureSessionInternal.ts
```diff
@@ -3,7 +3,26 @@
...
+	constructor(
+		public environment: Environment,
+		public userId: string,
+		public tenantId: string,
+		public accountInfo: AccountInfo | undefined,
+		private _credentials: AbstractCredentials | undefined,
+		public credentials2: AbstractCredentials2,
+	) {}
+
+	public get credentials(): AbstractCredentials {
```
ejizba, line 22: Not quite - we want the MSAL error inside the MsalAuthProvider, not in a generic class like this.

You need to delay calling `getCredentials` until it's actually used. Something like this:
```typescript
constructor(
    ...
    private _authProvider: AuthProviderBase<unknown>,
) { }

public get credentials(): AbstractCredentials {
    return this._authProvider.getCredentials(this.environment.name, this.userId, this.tenantId);
}

public get credentials2(): AbstractCredentials2 {
    return this._authProvider.getCredentials2(this.environment, this.userId, this.tenantId, this.accountInfo);
}
```

ejizba (approved):

wwlorey:

PR #310, opened 2021-09-09
Add 'Sign In with Device Code' for MSAL

wwlorey:
The device code authentication flow isn't supported for Microsoft employee accounts: https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/3951#issuecomment-915659897. But it works in cases where the "require managed device" grant control is disabled.

There was some confusion earlier when I thought I got this working for Microsoft employee accounts. I was using the client ID (configured with preauthorization) from the [device code sample](https://github.com/AzureAD/microsoft-authentication-library-for-js/tree/dev/samples/msal-node-samples/device-code) along with a pre-production authority (again, from the sample code). And under those conditions it works.


ejizba (approved):

wwlorey:

PR #312, opened 2021-09-09
Implement `loginWithoutLocalServer` for MSAL

wwlorey:


src/login/msal/MsalAuthProvider.ts
```diff
@@ -37,10 +37,6 @@ export class MsalAuthProvider extends AuthProviderBase<AuthenticationResult> {
 		this.publicClientApp = new PublicClientApplication(msalConfiguration);
 	}
 
-	public async loginWithoutLocalServer(_clientId: string, _environment: Environment, _isAdfs: boolean, _tenantId: string): Promise<AuthenticationResult> {
```
ejizba, line 40: Bit hard to tell - was this PR just moving code around or where there also changes specifically to make it work with msal?

src/login/AuthProviderBase.ts
```diff
@@ -24,8 +25,12 @@ export type AbstractCredentials2 = DeviceTokenCredentials2 | AzureIdentityCreden
 
 export abstract class AuthProviderBase<TLoginResult> {
 	private terminateServer: (() => Promise<void>) | undefined;
+	private handler: UriEventHandler = new UriEventHandler();
+
+	constructor() {
+		window.registerUriHandler(this.handler);
```
ejizba, line 31: Rather than move a bunch of code to this file, can we move it outside the auth provider classes? For example, you're now registering the uri handler twice (since we create an auth provider for both adal and msal), which at best is inefficient and at worst could cause bugs. It just doesn't seem like any of this code is unique to the auth providers

src/login/msal/MsalAuthProvider.ts
wwlorey: There were a few changes in addition to moving code:

### `loginWithoutLocalServer`
```diff
-	public async loginWithoutLocalServer(clientId: string, environment: Environment, isAdfs: boolean, tenantId: string): Promise<TokenResponse[]> {	
+	public async loginWithoutLocalServer(clientId: string, environment: Environment, isAdfs: boolean, tenantId: string): Promise<TLoginResult> {
		const callbackUri: Uri = await env.asExternalUri(Uri.parse(`${env.uriScheme}://ms-vscode.azure-account`));	
		const nonce: string = randomBytes(16).toString('base64');	
		const port: string | number = (callbackUri.authority.match(/:([0-9]*)$/) || [])[1] || (callbackUri.scheme === 'https' ? 443 : 80);	
		const callbackEnvironment: string = getCallbackEnvironment(callbackUri);	
		const state: string = `${callbackEnvironment}${port},${encodeURIComponent(nonce)},${encodeURIComponent(callbackUri.query)}`;	
		const signInUrl: string = `${environment.activeDirectoryEndpointUrl}${isAdfs ? '' : `${tenantId}/`}oauth2/authorize`;	
		let uri: Uri = Uri.parse(signInUrl);	
		uri = uri.with({	
			query: `response_type=code&client_id=${encodeURIComponent(clientId)}&redirect_uri=${redirectUrlAAD}&state=${state}&resource=${environment.activeDirectoryResourceId}&prompt=select_account`	
		});	
		void env.openExternal(uri);	
-		const timeoutPromise = new Promise((_resolve: (value: TokenResponse) => void, reject) => {	
+		const timeoutPromise = new Promise((_resolve: (value: TLoginResult) => void, reject) => {
			const wait = setTimeout(() => {	
				clearTimeout(wait);	
				reject('Login timed out.');	
			}, 1000 * 60 * 5)	
		});	
-		const tokenResponse: TokenResponse = await Promise.race([this.exchangeCodeForToken(clientId, environment, tenantId, redirectUrlAAD, state), timeoutPromise]);	
-		// eslint-disable-next-line @typescript-eslint/no-non-null-assertion	
-		await storeRefreshToken(environment, tokenResponse.refreshToken!);	
-		return getTokensFromToken(environment, tenantId, tokenResponse);	
+		return await Promise.race([this.exchangeCodeForToken(clientId, environment, tenantId, redirectUrlAAD, state), timeoutPromise]);
	}	

```

### `exchangeCodeForToken`
```diff
-	private async exchangeCodeForToken(clientId: string, environment: Environment, tenantId: string, callbackUri: string, state: string): Promise<TokenResponse> {
+	private async exchangeCodeForToken(clientId: string, environment: Environment, tenantId: string, callbackUri: string, state: string): Promise<TLoginResult> {
		let uriEventListener: Disposable;
-		return new Promise((resolve: (value: TokenResponse) => void , reject) => {
+ 		return new Promise((resolve: (value: TLoginResult) => void , reject) => {
			uriEventListener = this.handler.event(async (uri: Uri) => {
				try {
					/* eslint-disable @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access */
					const query = parseQuery(uri);
					const code = query.code;

					// Workaround double encoding issues of state
					if (query.state !== state && decodeURIComponent(query.state) !== state) {
						throw new Error('State does not match.');
					}
					/* eslint-enable @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access */

-					resolve(await getTokenWithAuthorizationCode(clientId, environment, callbackUri, tenantId, code));
+					resolve(await this.loginWithAuthCode(code, callbackUri, clientId, environment, tenantId));
				} catch (err) {
					reject(err);
				}
			});
		}).then(result => {
			uriEventListener.dispose()
			return result;
		}).catch(err => {
			uriEventListener.dispose();
			throw err;
		});
	}
```

src/login/AuthProviderBase.ts
wwlorey: Right now we're only [creating](https://github.com/microsoft/vscode-azure-account/blob/12ff3d1cdf34b511566565ccbf51fb1f086b8476/src/login/AzureLoginHelper.ts#L82-L84) one auth provider at a time but I agree this code might not belong here - I'll move it.

src/login/exchangeCodeForToken.ts
```diff
@@ -0,0 +1,56 @@
...
+import { AuthProviderBase } from "./AuthProviderBase";
+
+class UriEventHandler extends EventEmitter<Uri> implements UriHandler {
+	public handleUri(uri: Uri): void {
+		this.fire(uri);
+	}
+}
+
+const handler: UriEventHandler = new UriEventHandler();
+window.registerUriHandler(handler);
```
ejizba, line 17: I don't want us executing any code at the file-level like this. There's a risk of timing issues and any errors that occur would not be handled well (no telemetry, etc.). Instead, it should happen within the `activate` method or at least part of the `activate` call stack

Also, technically the result of `registerUriHandler` should be added to `ext.context.subscriptions` since it's a Disposable.

src/login/exchangeCodeForToken.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export async function exchangeCodeForToken<TLoginResult>(authProvider: AuthProviderBase<TLoginResult>, clientId: string, environment: Environment, tenantId: string, callbackUri: string, state: string): Promise<TLoginResult> {
```
ejizba, line 19: This is not exactly what I meant in terms of refactoring the files, but it seems like the problem is too big to solve in this PR anyways. Fix the uri handler stuff and that should be good for this PR

I don't have a suggested fix, but at a high level my concerns with the file structure are:
- Its hard to know the differences between all the "login" methods
- I don't understand the line between `AuthProviderBase` and `AzureLoginHelper`
- `AzureLoginHelper` is just such a giant mess of random code
- It's not clear which methods deserve their own file and which stay part of a class

ejizba (approved):

wwlorey:

PR #319, opened 2021-09-17
Add 'application' scope to a few settings

wwlorey:
Per an ask from the VS Code team

### DIFF
--- a/package.json
+++ b/package.json
@@ -141,7 +141,8 @@
         },
         "azure.customCloud.resourceManagerEndpointUrl": {
           "type": "string",
-          "description": "The management endpoint for your Azure Custom Cloud environment."
+          "description": "The management endpoint for your Azure Custom Cloud environment.",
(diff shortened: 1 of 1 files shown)


wwlorey:
More context on the application scope: 

> application - Settings that apply to all instances of VS Code and can only be configured in user settings. ([source](https://code.visualstudio.com/api/references/contribution-points))

neelip (approved):
Why only these two properties are updated? Wouldn't other properties in this file suit "application" scope as well?

wwlorey:
> Why only these two properties are updated? Wouldn't other properties in this file suit "application" scope as well?

These were the only changes requested by the VS Code team. None of the other settings immediately jump out at me for the "application" scope. But maybe it's worth going through these settings (and settings in other extensions) to evaluate which scopes are appropriate. Not very high priority tho.

Issue #323, opened 2021-09-21
Use "SecretStorage" VS Code API instead of keytar


wwlorey:
It makes sense to do this when we release MSAL since we'll have to sign people out anyway.

PR #329, opened 2021-09-30
Replace keytar with VS Code secrets API

wwlorey:
The secrets API was added in VS Code version 1.53.0: https://code.visualstudio.com/updates/v1_53#_secrets-api

As discussed offline: onboarding MSAL will require that all users sign out and back in. So now is a good time to move away from keytar since this change will also force users to sign in again. 

Fixes https://github.com/microsoft/vscode-azure-account/issues/323


bwateratmsft (approved):

bwateratmsft:
It's probably worth mentioning that we still get a fair number of those "org.freedesktop.secrets" issues with the secrets API in VSCode...since it uses `keytar` under the hood. Nothing for us to fix, but worth mentioning...

PR #331, opened 2021-09-30
Prep for 0.9.9

wwlorey:
Since CI/CD isn't set up this far back in the commit history, I grabbed the third party notice for the main branch. It includes `vscode-azureextensionui` (the only package added for this release) but it also includes a notice for `@azure/msal-node` which isn't included as a dependency here. Is that alright?


wwlorey:
Closing since this is targeting the wrong branch

PR #334, opened 2021-10-01
Add sort pre-commit hook

alexweininger:
This one is larger because I also converted the indentation for both files to 4 spaces to be consistent with our other projects. (also because the script does not respect indentation and defaults to 4 spaces)


wwlorey (approved):
When I run `npm install` locally with these changes it also updates the indentation in package-lock.json to be 4 spaces. Did you by chance run `npm install` before changing the indentation in package.json? Or is there something funky going on on my end?

alexweininger:
Yeah I think I ran `npm install` before changing the indentation in the package.json.

PR #345, opened 2021-10-28
Use the UI package for all telemetry

wwlorey:
@fiveisprime I'm seeing two machines with the `abexp.assignmentcontext` property in kusto now so I'm reasonably confident that the exp framework is working properly (there just seems to be a delay in picking up new data).

I updated all the extension's telemetry to use the UI package as a last-ditch effort to figure out why the exp framework properties weren't showing up but it didn't seem to be related to a new machine appearing. Anyways - having the exp framework setup for all telemetry events will be nice for getting some A/A tests going.


wwlorey:
> I thought for sure it would be fairly quick.

Yeah I thought the same thing.. I was talking with Jingaling yesterday about it and it sounds like that data is coming from queries intermittently run on the live cosmos DB which would explain why there's a bit of a delay.

Issue #356, opened 2021-11-11
Cloud shell fails to open on Mac


wwlorey:
This affects the current released version of the extension (0.9.10) and VS Code versions >= 1.62.

Attempting to open the cloud shell immediately results in this error notification:

![The terminal process "/Users/wilorey/.vscode-insiders/extensions/ms-vscode.azure-account-0.9.10/bin/node.sh '/Applications/Visual Studio Code - Insiders.app/Contents/Frameworks/Code - Insiders Helper (Renderer).app/Contents/MacOS/Code - Insiders Helper (Renderer)', '-e', 'require('/Users/wilorey/.vscode-insiders/extensions/ms-vscode.azure-account-0.9.10/dist/cloudConsoleLauncher').main()'" terminated with exit code: 1.](https://user-images.githubusercontent.com/22795803/141217421-8a68af0d-ad07-4cc4-8883-e80ac9770c58.png)

When I manually run the terminal process referenced in the above notification in VS Code v1.61 I get this error:
```bash
(node:27195) UnhandledPromiseRejectionWarning: Error: connect ECONNREFUSED 127.0.0.1:80
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1146:16)
(Use `Code Helper (Renderer) --trace-warnings ...` to show where the warning was created)
(node:27195) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:27195) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
Error: connect ECONNREFUSED 127.0.0.1:80
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1146:16) {
  errno: -61,
  code: 'ECONNREFUSED',
  syscall: 'connect',
  address: '127.0.0.1',
  port: 80
}
```

And if I run the same process in VS Code v1.62 the process ends silently with exit code 1.

It appears that the future came and that unhandled promise rejections now cause the node process to end. I have yet to track down the culprit of that unhandled rejection.

Also @nturinski and I both verified that this error doesn't repro on Windows.

[...]

PR #357, opened 2021-11-12
Pass in `--ms-enable-electron-run-as-node` flag when creating the cloud shell process

wwlorey:
See [this comment](https://github.com/microsoft/vscode/issues/136987#issuecomment-966914578) for more context on why this change is needed. The `--ms-enable-electron-run-as-node` flag isn't required on Windows and including it throws an error. So I only add the flag for Mac & Linux.

Fixes https://github.com/microsoft/vscode-azure-account/issues/356


alexweininger (approved):
Nice. I verified the fix works for me on macOS.

Issue #354, opened 2021-11-06
Show help notification if no subscriptions found


wwlorey:
It's a common problem that no subscriptions are found when attempting to list subscriptions. In many cases this is because someone needs to set their tenant using the `azure.tenant` setting, but that isn't very discoverable.

In the help notification, we should link to a wiki page or some form of documentation with instructions on how to set the tenant.

wwlorey:
In addition to instructions for dealing with known problems related to setting the tenant, we should point people to a GitHub issue template to gather more info specifically related to this "no subscriptions found" problem. Hopefully a targeted issue template can help us understand the root cause of this problem if the tenant fix doesn't work.

PR #362, opened 2021-11-18
Provide a link to troubleshooting page when no subscriptions are found

wwlorey:
Fixes https://github.com/microsoft/vscode-azure-account/issues/354

### DIFF
--- a/src/login/AzureLoginHelper.ts
+++ b/src/login/AzureLoginHelper.ts
@@ -412,14 +412,14 @@ export class AzureLoginHelper {
 	}
 
 	private async showNoSubscriptionsFoundNotification(): Promise<void> {
-		const noSubscriptionsFound = localize('azure-account.noSubscriptionsFound', 'No subscriptions were found. Check out our troubleshooting page for common solutions to this problem or setup your account.');
-		const openTroubleshooting = localize('azure-account.openTroubleshooting', 'Open Troubleshooting');
+		const noSubscriptionsFound = localize('azure-account.noSubscriptionsFound', 'No subscriptions were found. Setup your account if you have yet to do so or check out our troubleshooting page for common solutions to this problem.');
 		const setupAccount = localize('azure-account.setupAccount', 'Setup Account');
-		const response = await window.showInformationMessage(noSubscriptionsFound, openTroubleshooting, setupAccount);
-		if (response === openTroubleshooting) {
+		const openTroubleshooting = localize('azure-account.openTroubleshooting', 'Open Troubleshooting');
+		const response = await window.showInformationMessage(noSubscriptionsFound, setupAccount, openTroubleshooting);
+		if (response === setupAccount) {
+			void openUri('https://aka.ms/AAeyf8k');
+		} else if (response === openTroubleshooting) {
 			void openUri('https://aka.ms/AAevvhr');
-		} else if (response === setupAccount) {
-			void openUri('https://aka.ms/AAevntl');
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


src/login/AzureLoginHelper.ts
```diff
@@ -411,11 +411,15 @@ export class AzureLoginHelper {
 		});
 	}
 
-	private async noSubscriptionsFound(): Promise<void> {
-		const open: MessageItem = { title: localize('azure-account.open', "Open") };
-		const response: MessageItem | undefined = await window.showInformationMessage(localize('azure-account.noSubscriptionsFound', "No subscriptions were found. Set up your account at https://azure.microsoft.com/en-us/free/."), open);
-		if (response === open) {
-			void openUri('https://azure.microsoft.com/en-us/free/?utm_source=campaign&utm_campaign=vscode-azure-account&mktingSource=vscode-azure-account');
+	private async showNoSubscriptionsFoundNotification(): Promise<void> {
+		const noSubscriptionsFound = localize('azure-account.noSubscriptionsFound', 'No subscriptions were found. Check out our troubleshooting page for common solutions to this problem or setup your account.');
```
bwateratmsft, line 415: Should we present setup as the first option or troubleshooting? I can see valid arguments both ways. @fiveisprime what do you think?

Either way, we should make the button order match the prose (as you have here).

bwateratmsft (approved):

src/login/AzureLoginHelper.ts
wwlorey: > Should we present setup as the first option or troubleshooting? I can see valid arguments both ways.

I agree - I could see it going either way. I'm cool with deferring to @fiveisprime on this

alexweininger (approved):

src/login/AzureLoginHelper.ts
wwlorey: @jinglouMSFT, @alexweininger, and I put on our PM hats and decided that it's better if we swap the message's order. Also, both links now take you to different headers in the [troubleshooting page](https://github.com/microsoft/vscode-azure-account/wiki/Troubleshooting) where I added some more details about setting up your account.

src/login/AzureLoginHelper.ts
```diff
@@ -411,11 +411,15 @@ export class AzureLoginHelper {
...
-		const open: MessageItem = { title: localize('azure-account.open', "Open") };
-		const response: MessageItem | undefined = await window.showInformationMessage(localize('azure-account.noSubscriptionsFound', "No subscriptions were found. Set up your account at https://azure.microsoft.com/en-us/free/."), open);
-		if (response === open) {
-			void openUri('https://azure.microsoft.com/en-us/free/?utm_source=campaign&utm_campaign=vscode-azure-account&mktingSource=vscode-azure-account');
+	private async showNoSubscriptionsFoundNotification(): Promise<void> {
+		const noSubscriptionsFound = localize('azure-account.noSubscriptionsFound', 'No subscriptions were found. Setup your account if you have yet to do so or check out our troubleshooting page for common solutions to this problem.');
+		const setupAccount = localize('azure-account.setupAccount', 'Setup Account');
+		const openTroubleshooting = localize('azure-account.openTroubleshooting', 'Open Troubleshooting');
+		const response = await window.showInformationMessage(noSubscriptionsFound, setupAccount, openTroubleshooting);
+		if (response === setupAccount) {
```
nturinski, line 419: I think adding telemetry to this (if possible, not sure if the telemetryReporter is set up yet) would be useful.  It'd be a nice way to tell how many people are running into errors because they just don't have accounts or have an account that isn't showing up.

Issue #365, opened 2021-11-23
Throw an error if users attempt to use Cloud Shell while signed in with MSAL


wwlorey:
The Cloud Shell team is still investigating the ICM I filed about using MSAL with Cloud Shell (more details in https://github.com/microsoft/vscode-azure-account/issues/313#issuecomment-970929428). So we're not going to ship MSAL support for Cloud Shell in v0.10.0.

PR #368, opened 2021-11-24
Throw a pretty error when attempting to authenticate Cloud Shell with MSAL

wwlorey:
Fixes https://github.com/microsoft/vscode-azure-account/issues/365

I opted to add a log to the shell instead of throwing a "traditional" error because the TerminalProfile throws additional errors when a profile isn't provided (which is what happens when we throw an error from `createCloudConsole`)


wwlorey:
Here's what the error looks like in the shell:

Screen Shot 2021-11-23 at 4 24 19 PM

bwateratmsft (approved):

Issue #328, opened 2021-09-30
Consider restructuring login files


wwlorey:
There's some engineering work related to the login code that would be nice to have:

> * Its hard to know the differences between all the "login" methods
> * I don't understand the line between AuthProviderBase and AzureLoginHelper
> * AzureLoginHelper is just such a giant mess of random code
> * It's not clear which methods deserve their own file and which stay part of a class

(From https://github.com/microsoft/vscode-azure-account/pull/312#discussion_r706403843)

PR #370, opened 2021-11-24
Dissect AzureLoginHelper.ts into multiple files

wwlorey:
Fixes https://github.com/microsoft/vscode-azure-account/issues/328

AzureLoginHelper.ts had become pretty bloated so I moved many of its functions to their own files and created a "commands" folder. I also renamed `AzureLoginHelper` to `AzureAccountLoginHelper` to hopefully convey that it's the main login helper for the "AzureAccount" extension and that `AuthProviderBase`, `AdalAuthProvider`, and `MsalAuthProvider` are subordinate to it.


src/login/AzureLoginHelper.ts
```diff
@@ -171,14 +139,10 @@ export class AzureLoginHelper {
 		});
 	}
 
-	private async sendLoginTelemetry(context: IActionContext, trigger: LoginTrigger, path: CodePath, cloud: string, outcome: string, message?: string, includeSubscriptions?: boolean) {
+	private async sendLoginTelemetry(context: IActionContext, properties: { trigger: LoginTrigger, codePath: CodePath, environmentName: string, outcome: string, message?: string }, includeSubscriptions?: boolean) {
```
wwlorey, line 142: I've been meaning to do this for a while: it's bad practice to have a bunch of strings in sequence for function arguments because they can be easily swapped so this makes it more robust & the code is cleaner.

src/login/commands/selectSubscriptions.ts
```diff
@@ -0,0 +1,101 @@
...
+	const resourceFilterConfig = azureConfig.inspect<string[]>(resourceFilterSetting);
+	const target: ConfigurationTarget = getCurrentTarget(resourceFilterConfig);
+	await azureConfig.update(resourceFilterSetting, resourceFilter[0] !== 'all' ? resourceFilter : undefined, target);
+}
+
+function showNoSubscriptionsFoundNotification(): void {
+	const noSubscriptionsFound = localize('azure-account.noSubscriptionsFound', 'No subscriptions were found. Setup your account if you have yet to do so or check out our troubleshooting page for common solutions to this problem.');
+	const setupAccount = localize('azure-account.setupAccount', 'Setup Account');
+	const openTroubleshooting = localize('azure-account.openTroubleshooting', 'Open Troubleshooting');
+	void window.showInformationMessage(noSubscriptionsFound, setupAccount, openTroubleshooting).then(response => {
```
wwlorey, line 94: When I implemented this I forgot that we shouldn't await UI because it blocks telemetry, so we no longer await this call

wwlorey (commented):
All of the changes made here are just moving code around except for the two cases that I've highlighted

src/login/AzureLoginHelper.ts
bwateratmsft: Other than this are there any substantive changes or is it just splitting it out into multiple files?

bwateratmsft (approved):

alexweininger (approved):

src/login/AzureLoginHelper.ts
wwlorey: > Other than this are there any substantive changes or is it just splitting it out into multiple files?

Nope these were the only substantive changes. It's mainly just moving a lot of code around

Issue #371, opened 2021-11-25
[Suggestion] It would be better if adding a "Azure Account: Report Issue..." command in the command palette


v-mengwe:
**OS:** All
**VS Code Version:** 1.62.3
**Build Version:** 20211124.8

**Repro Steps:**

1. F1 -> Type "report issue".
2. Check the command palette.

**Actual:**
There is no "Azure Account: Report Issue..." command in the command palette.
![MicrosoftTeams-image (1)](https://user-images.githubusercontent.com/62053665/143409237-a233cd7f-0ab7-4690-8fab-82f80d12de72.png)

**Suggestion:**
It would be better if adding a "Azure Account: Report Issue..." command in the command palette.

PR #375, opened 2021-11-29
Add `Azure Account: Report Issue...` command

wwlorey:
Fixes https://github.com/microsoft/vscode-azure-account/issues/371

I added this command under the category "Azure Account" instead of "Azure" which is what we normally do for this extension. The `Azure: Report Issue...` command already exists and it would look gross if we added another command with the same name:

Screen Shot 2021-11-29 at 12 46 00 PM

### DIFF
--- a/package.nls.json
+++ b/package.nls.json
@@ -2,7 +2,7 @@
     "azure-account.cloudShellBash": "Azure Cloud Shell (Bash)",
     "azure-account.cloudShellPowerShell": "Azure Cloud Shell (PowerShell)",
     "azure-account.commands.azure": "Azure",
-    "azure-account.commands.azure-account": "Azure Account",
+    "azure-account.commands.azureAccount": "Azure Account",
     "azure-account.commands.createAccount": "Create an Account",
     "azure-account.commands.login": "Sign In",
     "azure-account.commands.loginToCloud": "Sign In to Azure Cloud",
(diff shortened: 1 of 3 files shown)


package.nls.json
```diff
@@ -2,11 +2,13 @@
     "azure-account.cloudShellBash": "Azure Cloud Shell (Bash)",
     "azure-account.cloudShellPowerShell": "Azure Cloud Shell (PowerShell)",
     "azure-account.commands.azure": "Azure",
+    "azure-account.commands.azureAccount": "Azure Account",
```
bwateratmsft, line 5: Maybe `Azure Account Extension`? I'm worried it will be confused for a way for anybody with essentially any problems with their Azure account to report issues. @fiveisprime want to provide some PM input too?

bwateratmsft (approved):

package.nls.json
wwlorey: I do get the concern. I think worst case we can always direct people who file issues related to general Azure or Azure accounts to the right place

nturinski: Yeah, I think even if we added Azure Account Extension, we'll still issues filed against us for general Azure Account problems because I don't think most users will be able to distinguish when it's something that our extension is/is not doing vs. their Azure Account is just busted.

And even if it's a problem with their Azure Account, there may be steps that we can take in the extension to resolve/guide them as well.

PR #382, opened 2021-12-06
Set authentication library based on user setting and exp framework

wwlorey:
The user's `azure.authLibrary` setting takes precedence over the auth library the experimentation framework tells us to use. Also I added telemetry for the auth library setting so we can filter out users from the A/B test reports who manually set their auth library (I believe that's all that is required in the extension to do that filtering)

### DIFF
--- a/src/login/getAuthLibrary.ts
+++ b/src/login/getAuthLibrary.ts
@@ -9,8 +9,7 @@ import { ext } from "../extensionVariables";
 import { getSettingValue } from "../utils/settingUtils";
 
 export function getAuthLibrary(): AuthLibrary {
-	// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-	return (callWithTelemetryAndErrorHandlingSync('getAuthLibrary', (context: IActionContext) => {
+	return callWithTelemetryAndErrorHandlingSync('getAuthLibrary', (context: IActionContext) => {
 		let authLibrary: AuthLibrary | undefined = getSettingValue<AuthLibrary>(authLibrarySetting);
 
 		switch (authLibrary) {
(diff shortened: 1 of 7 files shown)


src/login/getAuthLibrary.ts
```diff
@@ -0,0 +1,34 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import { callWithTelemetryAndErrorHandlingSync, IActionContext } from "vscode-azureextensionui";
+import { AuthLibrary, authLibrarySetting } from "../constants";
+import { ext } from "../extensionVariables";
+import { getSettingValue } from "../utils/settingUtils";
+
+export function getAuthLibrary(): AuthLibrary {
+	// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+	return (callWithTelemetryAndErrorHandlingSync('getAuthLibrary', (context: IActionContext) => {
```
bwateratmsft, line 13: I don't know for sure without trying, but this might eliminate the need for the non-null assertion (and consequently the eslint suppression, and one pair of parentheses):
```suggestion
	return (callWithTelemetryAndErrorHandlingSync('getAuthLibrary', (context: IActionContext): AuthLibrary => {
```

Otherwise, `as AuthLibrary` below might work too.

src/login/getAuthLibrary.ts
```diff
@@ -0,0 +1,34 @@
...
+				context.telemetry.properties.authLibrarySetting = 'undefined';
+				if (ext.isMsalTreatmentVariable) {
+					authLibrary = 'MSAL';
+				} else {
+					authLibrary = 'ADAL';
+				}
+		}
+
+		return authLibrary;
+	}))!;
```
bwateratmsft, line 33: Can we split to a few lines with some bonus indentation? Having `}))!;` all at once is a little jarring 😨

Assuming the above type assertion does work I think it'll just be `});` which is fine, no need for additional whitespacing.

bwateratmsft (approved):

Issue #386, opened 2021-12-13
There is an error under the azure account registry when setting the "Azure: Authentication Library" setting to "MSAL"


v-mengwe:
**OS:** Windows 10
**Build Version:** 20211210.3
**Regression:** Not a regression

**Repro Steps:**

1. Set the "Azure: Authentication Library" setting to "MSAL".
2. Install the Docker extension in VS Code.
3. Expand the "REGISTRIES" node -> Create an azure account registry -> Expand the azure account registry.
4. Check whether there is no error under the azure account registry.

**Expect:**
There is no error under the azure account registry.

**Actual:**
There is an error under the azure account registry.
![image](https://user-images.githubusercontent.com/62053665/145786324-a796fe56-300d-4947-ae9a-dcc672b9859f.png)

**More Info:**
This issue doesn't reproduce when setting the "Azure: Authentication Library" setting to "ADAL"

bwateratmsft:
Looks like the MSAL credentials aren't implementing [`AzExtServiceClientCredentials`](https://github.com/microsoft/vscode-azuretools/blob/main/ui/index.d.ts#L147-L155) quite right...

PR #388, opened 2021-12-27
Implement `signRequest` in `PublicClientCredential`

wwlorey:
Fixes https://github.com/microsoft/vscode-azure-account/issues/386

### DIFF
--- a/src/login/msal/PublicClientCredential.ts
+++ b/src/login/msal/PublicClientCredential.ts
@@ -4,10 +4,12 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { AccessToken, TokenCredential } from "@azure/core-auth";
+import { Constants as MSRestConstants, WebResource } from "@azure/ms-rest-js";
 import { AccountInfo, AuthenticationResult, PublicClientApplication } from "@azure/msal-node";
+import { AzExtServiceClientCredentials } from "vscode-azureextensionui";
 import { defaultMsalScopes } from "../../constants";
 
-export class PublicClientCredential implements TokenCredential {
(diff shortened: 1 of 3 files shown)


src/login/msal/PublicClientCredential.ts
```diff
@@ -38,4 +40,15 @@ export class PublicClientCredential implements TokenCredential {
 
 		return null;
 	}
+
+	public async signRequest(webResource: WebResource): Promise<WebResource | undefined> {
+		const tokenResponse: AccessToken | null = await this.getToken(defaultMsalScopes);
+		if (tokenResponse) {
+			webResource.headers.set(
+				MSRestConstants.HeaderConstants.AUTHORIZATION,
+				`${MSRestConstants.HeaderConstants.AUTHORIZATION_SCHEME} ${tokenResponse.token}`
```
bwateratmsft, line 49: Is the auth scheme part of the token response or will it always be fixed?

bwateratmsft (approved):

src/login/msal/PublicClientCredential.ts
wwlorey: The scheme is fixed - it'll always be "Bearer". Would it be better to hard code that instead of using the MS Rest constant?

bwateratmsft: I think it's fine to use the constant. I've seen in some places that response to the token request includes the type in addition to the token--so, it could be e.g. `Authorization: Bearer abcd123` or `Authorization: JWT 123abcd`, with the token response including the type `Bearer` or `JWT`. But if the token response has no such field, I'd say what you have is the way to go.

Issue #101, opened 2018-11-05
Add "Select Tenant" command


chrmarti:

kilasuit:
Any updates on this @chrmarti 

As a consultant with access to many tenants I do not want my tooling to be opening access to all tenants at a time even if I am using the same AAD account to login to each tenant with.

kilasuit:
Whilst I can get round this with the following in my settings.json file this only forces me into the current set tenant & and is not optimal for switching between tenants. 

```json
"azure.tenant": ""
```

RMacfarlane:
@chrmarti Was the idea with this that we would show a quickpick with available tenants instead of having the user go find the tenant id themselves?

chrmarti:
@RMacfarlane Yes, you could also allow for a custom tenant id because when listing tenants we don't necessarily get the complete list of tenants the user has access to. Not sure how important that is though, so having a QuickPick to choose from the known tenant ids might be the main improvement here.

wwlorey:
This could help us set users back on track who are failing to list any of their subscriptions.

`SubscriptionClient` gives us the ability to list known tenants and we can aggregate those into the quick pick. Also, we should replace our guidance on the wiki for manually finding tenant IDs and direct users to use this command instead.

wwlorey:
We shipped support for this command in [v0.10.0](https://github.com/microsoft/vscode-azure-account/releases/tag/v0.10.0).

PR #395, opened 2022-01-06
Add 'Select Tenant' command

wwlorey:
Fixes https://github.com/microsoft/vscode-azure-account/issues/101

I also updated the [troubleshooting wiki page](https://github.com/microsoft/vscode-azure-account/wiki/Troubleshooting) to recommend this command first when troubleshooting tenant problems.

One potential issue here is if a user enters a custom tenant that is wrong then attempts to sign in, they won't be able to use the "Select Tenant" command to update the tenant. This is because we require the user to be signed in when this command is run to dynamically get a list of available tenants.

### DIFF
--- a/src/login/commands/selectTenant.ts
+++ b/src/login/commands/selectTenant.ts
@@ -1,0 +1,67 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { commands, window } from "vscode";
+import { callWithTelemetryAndErrorHandling, IActionContext } from "vscode-azureextensionui";
+import { tenantSetting } from "../../constants";
+import { ext } from "../../extensionVariables";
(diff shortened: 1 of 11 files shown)


src/login/commands/selectTenant.ts
```diff
@@ -0,0 +1,67 @@
...
+
+		if (!tenants.length) {
+			context.telemetry.properties.outcome = 'noTenantsFound';
+			const noTenantsFound = localize('azure-account.noTenantsFound', 'No tenants were found for the subscription(s) you\'ve selected.');
+			const learnMoreAboutTenants = { title: localize('azure-account.learnMoreAboutTenants', 'Learn more about tenants') };
+			void context.ui.showWarningMessage(noTenantsFound, learnMoreAboutTenants).then(result => {
+				if (result === learnMoreAboutTenants) {
+					void openUri('https://aka.ms/AAfe10l');
+				}
+			});
```
wwlorey, line 38: I'm open to input as to whether or not this is needed... It seems like a nice addition but I'm not sure how often this case will happen. Also since we provide the option to enter a custom tenant the quickpick list will ever be empty.

bwateratmsft (approved):

PR #397, opened 2022-01-11
Rework 'Select Tenant' command

wwlorey:
I didn't properly understand the relationship between tenants & subscriptions when I first introduced the "Select Tenant" command in https://github.com/microsoft/vscode-azure-account/pull/395 so I've fixed it here. Shoutout to @nturinski and @alexweininger for helping me get a better handle on how these things are related.

A few things to note about this change:

* We can only list subscriptions associated with the tenant the user signed in with. We have access to list all tenants available to a particular user though. So in the cache, the list of all available tenants is now stored at the same level in the object as the subscriptions list.
* We no longer require the user to be signed in to run the "Select Tenant" command. This requirement originally created a "chicken & the egg" problem. Now if they aren't signed in, only the "Enter custom tenant" option is shown (no auto-fetched tenants). This means that the "No tenants found" notification needed to be removed since that's a mainline scenario now.

### DIFF
--- a/src/login/AzureLoginHelper.ts
+++ b/src/login/AzureLoginHelper.ts
@@ -6,7 +6,7 @@
 import { Environment } from '@azure/ms-rest-azure-env';
 import { CancellationTokenSource, commands, EventEmitter, ExtensionContext, MessageItem, window, workspace } from 'vscode';
 import { callWithTelemetryAndErrorHandling, IActionContext } from 'vscode-azureextensionui';
-import { AzureLoginStatus, AzureSession } from '../azure-account.api';
+import { AzureLoginStatus, AzureResourceFilter, AzureSession, AzureSubscription } from '../azure-account.api';
 import { authLibrarySetting, cacheKey, clientId, cloudSetting, commonTenantId, customCloudArmUrlSetting, resourceFilterSetting, tenantSetting } from '../constants';
 import { AzureLoginError, getErrorMessage } from '../errors';
 import { ext } from '../extensionVariables';
@@ -25,7 +25,7 @@ import { getAuthLibrary } from './getAuthLibrary';
 import { getKey } from './getKey';
 import { MsalAuthProvider } from './msal/MsalAuthProvider';
 import { checkRedirectServer } from './server';
-import { AzureResourceFilterInternal, AzureSubscriptionInternal, ISubscriptionCache } from './subscriptionTypes';
+import { SubscriptionTenantCache } from './subscriptionTypes';
 import { updateFilters } from './updateFilters';
 import { waitUntilOnline } from './waitUntilOnline';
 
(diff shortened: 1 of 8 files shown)


bwateratmsft (approved):

src/login/AzureLoginHelper.ts
```diff
@@ -189,27 +190,27 @@ export class AzureAccountLoginHelper {
...
 				};
 			});
 			this.subscriptionsTask = Promise.resolve(subscriptions);
 			this.api.subscriptions.push(...subscriptions);
+			this.tenantsTask = Promise.resolve(cache.tenants);
 
 			const resourceFilter: string[] | undefined = getSettingValue(resourceFilterSetting);
 			this.oldResourceFilter = JSON.stringify(resourceFilter);
-			const newFilters: AzureResourceFilterInternal[] = getNewFilters(subscriptions, resourceFilter);
+			const newFilters: AzureSubscription[] = getNewFilters(subscriptions, resourceFilter);
```
nturinski, line 213: Is this supposed to be `AzureSubscription[]`? The return type of `getNewFilters` is `AzureResourceFilter[]`.  Does `AzureSubscription` extend it?

src/login/subscriptionTypes.ts
```diff
@@ -12,7 +12,7 @@ export interface ISubscriptionItem extends QuickPickItem {
 	subscription: AzureSubscription;
 }
 
-export interface ISubscriptionCache {
+export interface SubscriptionTenantCache {
```
nturinski, line 15: Soooo... we done with the I prefixes then? 😗

src/login/updateFilters.ts
```diff
@@ -16,11 +16,11 @@ export function updateFilters(configChange = false): void {
 	}
 	ext.loginHelper.filtersTask = (async () => {
 		await ext.loginHelper.api.waitForSubscriptions();
-		const subscriptions: AzureSubscriptionInternal[] = await ext.loginHelper.subscriptionsTask;
+		const subscriptions: AzureSubscription[] = await ext.loginHelper.subscriptionsTask;
 		ext.loginHelper.oldResourceFilter = JSON.stringify(resourceFilter);
-		const newFilters: AzureResourceFilterInternal[] = getNewFilters(subscriptions, resourceFilter);
+		const newFilters: AzureSubscription[] = getNewFilters(subscriptions, resourceFilter);
```
nturinski, line 21: Here too?

src/login/AzureLoginHelper.ts
wwlorey: Yessir they're actually the same type:

https://github.com/microsoft/vscode-azure-account/blob/9b8a76ddc8f9ce181f71fdf283b1d23009ca37c2/src/azure-account.api.d.ts#L47

This as well as the other instance you mentioned should technically be `AzureResourceFilter` but didn't wanna crowd the PR with those changes.

src/login/subscriptionTypes.ts
wwlorey: 😁😁  I decided to hop on the "no I" train but we should probs be consistent... Should we discuss next standup?

src/login/AzureLoginHelper.ts
wwlorey: Took care of this here: https://github.com/microsoft/vscode-azure-account/pull/400 💯💯💯💯

src/login/subscriptionTypes.ts
nturinski: lol no, I never want to have another 2 hour conversation about it again.  Let's just not use I moving forward

PR #402, opened 2022-01-14
Use vscode-azureextensiondev for webpacking

wwlorey:
Fixes https://github.com/microsoft/vscode-azure-account/issues/392

The above issue happened because the "open" package (which is used by the UI package to open the "Report Issue" link) wasn't configured properly in this extension for webpack. "open" is configured in "vscode-azureextensiondev" already so this was a good excuse for switching the whole extension over to using the dev package for webpacking.


src/utils/getCoreNodeModule.ts
```diff
@@ -0,0 +1,26 @@
...
+
+import * as vscode from 'vscode';
+
+/**
+ * Returns a node module installed with VSCode, or undefined if it fails.
+ */
+export function getCoreNodeModule<T>(moduleName: string): T | undefined {
+    try {
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+        return require(`${vscode.env.appRoot}/node_modules.asar/${moduleName}`);
```
alexweininger, line 14: Could you explain what `node_modules.asar` is and why some modules come from there? I couldn't find code like this in any other webpack we have, so I'm guessing this is something account specific.

alexweininger (approved):

src/utils/getCoreNodeModule.ts
wwlorey: In case you aren't familiar (I was not lol), asar is an archive format. Here's a related node module that explains it a bit: https://www.npmjs.com/package/asar.

This `getCoreNodeModule` code is present in other extensions (at least [storage](https://github.com/microsoft/vscode-azurestorage/blob/a0abd4fbcfcf276f626c3c1cdc270ba9b5c98aa1/src/utils/getCoreNodeModule.ts#L11) and [functions](https://github.com/microsoft/vscode-azurefunctions/blob/15768d49fdf2118cc900c612160c0e3fb8b3a164/src/utils/getCoreNodeModule.ts#L11)). As for why we expect some modules to come from `node_modules.asar` and how that version of node modules would get ever created I'm not sure. But I left it in because this is a pretty classic [Chesterton’s Fence](https://wiki.lesswrong.com/wiki/Chesterton%27s_Fence) situation.

Issue #415, opened 2022-01-27
Investigate supporting T2 credentials for ADAL


wwlorey:
Right now we only support T1 for ADAL and T2 for MSAL. But we don't want to force users to switch to MSAL for them to use the new T2 SDK.

bwateratmsft:
In the one use case I found that Docker is using those credentials, we could use `getToken`. In fact that'd probably be easier.

Would fixing this issue be a prerequisite to unconditionally using `getToken` regardless of MSAL vs. ADAL?

bwateratmsft:
(Bottom line is I'm trying to make sure the Docker extension doesn't get busted for this, regardless of Azure Account extension version, or ADAL vs. MSAL)

nturinski:
> Would fixing this issue be a prerequisite to unconditionally using getToken regardless of MSAL vs. ADAL?

I think either way should be okay for Docker.  Though `signRequest` should definitely work since the credentials haven't changed since MSAL tokens appear to work for clients that require `signRequest`.  The issue we're running into is having an invalid access token when using ADAL with the new `WebsiteManagementClient`.  

I think if we implement `getToken` on the old credentials, it could be forward compatible.

wwlorey:
Some thoughts @nturinski and I had after discussing offline: 

We should try getting rid of `DeviceTokenCredentials` for ADAL and replace it with our own class implementing `TokenCredential` and [`AzExtServiceClientCredentials`](https://github.com/microsoft/vscode-azuretools/blob/main/ui/index.d.ts#L147) (similar to how we created [`PublicClientCredential`](https://github.com/microsoft/vscode-azure-account/blob/30056757606d4fa73f38b9b521f14c34d7455b11/src/login/msal/PublicClientCredential.ts#L12)). We will need to verify these typings are compatible with type 1 and type 2 credentials.

PR #418, opened 2022-02-01
Add a forward compatible token to use for T1&T2 sdks

nturinski:
So full disclosure... Not _SUPER_ confident about this approach, but some context for this problem:

Fixes https://github.com/microsoft/vscode-azure-account/issues/415#issuecomment-1024795447

On Friday, Will and I investigated why the T1 and T2 sdks weren't compatible.  It turns out it mostly has to do with the `getToken` function built into `DeviceTokenCredentials`.  It's return type is `TokenResponse` instead of `AccessToken` like it is for `TokenCredential` which is used by the T2 sdk.

AFAIK, T1 clients only require `ServiceClientCredentials` which only has  an implementation of `signRequest` not `getToken`.  I'm not sure why the Account Extension is architected to use `DeviceTokenCredentials`, but there are a lot of properties on it that I would not mess with and break everything.

I thought the simplest approach would be to just create a new class that overwrites the two methods that we care about, `getToken` and `signRequest` and implement them in a way that is compatible with both T1 and T2 SDKs.  

The reason we figured that this would be okay is because the MSAL tokens that we have been testing have been working perfectly fine with all of our T1 packages up to today which means the tokens seem backwards compatible.

Though I'm a bit nervous because I don't know how `getToken` interacts with other packages, but in the Azure Account extension itself, I don't see it being called anywhere.  With my limited testing, nothing seemed to be broken and both packages work.

I had to overwrite `signRequest` because the old implementation used `getToken` which expected a `TokenResponse` rather than `AccessToken`.  It was an easy fix, but my confidence level that this doesn't break other things isn't super high.

I'm open to any other suggestion, honestly.

### DIFF
--- a/src/login/adal/DeviceTokenCredentials2.ts
+++ b/src/login/adal/DeviceTokenCredentials2.ts
@@ -4,7 +4,6 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { AccessToken, TokenCredential } from "@azure/core-auth";
-import { Constants as MSRestConstants, WebResource } from "@azure/ms-rest-js";
 import { DeviceTokenCredentials } from '@azure/ms-rest-nodeauth';
 import { TokenResponse } from "adal-node";
 
@@ -16,18 +15,8 @@ import { TokenResponse } from "adal-node";
  * overwrite that implementation
  */
 export class DeviceTokenCredentials2 extends DeviceTokenCredentials implements TokenCredential {
-	// eslint-disable-next-line @typescript-eslint/no-explicit-any
-	public async getToken(): Promise<any> {
-		const tokenResponse = <AccessToken & TokenResponse>(await this.getTokenFromCache(this.username));
+	public async getToken(): Promise<AccessToken & TokenResponse> {
+		const tokenResponse = await super.getToken();
 		return Object.assign(tokenResponse, {token: tokenResponse.accessToken, expiresOnTimestamp: tokenResponse.expiresIn});
-	}
-
-	public async signRequest(webResource: WebResource): Promise<WebResource> {
-		const tokenResponse: AccessToken = <AccessToken>(await this.getToken());
-			webResource.headers.set(
-				MSRestConstants.HeaderConstants.AUTHORIZATION,
-				`${MSRestConstants.HeaderConstants.AUTHORIZATION_SCHEME} ${tokenResponse.token}`
-			);
-		return webResource;
 	}
 }
--- a/src/login/adal/DeviceTokenCredentials2.ts
+++ b/src/login/adal/DeviceTokenCredentials2.ts
@@ -14,7 +14,7 @@ import { DeviceTokenCredentials } from '@azure/ms-rest-nodeauth';
  * `DeviceTokenCredentials` forces `getToken` to return `TokenResponse` so this is to
  * overwrite that implementation
  */
-export class ForwardCompatibleToken extends DeviceTokenCredentials implements TokenCredential {
+export class DeviceTokenCredentials2 extends DeviceTokenCredentials implements TokenCredential {
 	// eslint-disable-next-line @typescript-eslint/no-explicit-any
 	public async getToken(): Promise<any> {
 		const tokenResponse = await this.getTokenFromCache(this.username);
--- a/src/login/adal/DeviceTokenCredentials2.ts
+++ b/src/login/adal/DeviceTokenCredentials2.ts
@@ -6,7 +6,6 @@
 import { AccessToken, TokenCredential } from "@azure/core-auth";
 import { Constants as MSRestConstants, WebResource } from "@azure/ms-rest-js";
 import { DeviceTokenCredentials } from '@azure/ms-rest-nodeauth';
-import { TokenResponse } from "adal-node";
 
 /**
  * Token that is forward compatible with track 2 Azure SDK for Node.js
@@ -18,11 +17,12 @@ import { TokenResponse } from "adal-node";
 export class DeviceTokenCredentials2 extends DeviceTokenCredentials implements TokenCredential {
 	// eslint-disable-next-line @typescript-eslint/no-explicit-any
 	public async getToken(): Promise<any> {
-		const tokenResponse = <AccessToken & TokenResponse>(await this.getTokenFromCache(this.username));
-		return Object.assign(tokenResponse, {token: tokenResponse.accessToken, expiresOnTimestamp: tokenResponse.expiresIn});
+		const tokenResponse = await this.getTokenFromCache(this.username);
+		return <AccessToken>{ token: tokenResponse.accessToken, expiresOnTimestamp: tokenResponse.expiresIn };
+
 	}
 
-    public async signRequest(webResource: WebResource): Promise<WebResource> {
+	public async signRequest(webResource: WebResource): Promise<WebResource> {
 		const tokenResponse: AccessToken = <AccessToken>(await this.getToken());
 			webResource.headers.set(
 				MSRestConstants.HeaderConstants.AUTHORIZATION,
--- a/src/login/adal/DeviceTokenCredentials2.ts
+++ b/src/login/adal/DeviceTokenCredentials2.ts
@@ -6,6 +6,7 @@
 import { AccessToken, TokenCredential } from "@azure/core-auth";
 import { Constants as MSRestConstants, WebResource } from "@azure/ms-rest-js";
 import { DeviceTokenCredentials } from '@azure/ms-rest-nodeauth';
+import { TokenResponse } from "adal-node";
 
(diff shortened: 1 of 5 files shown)


bwateratmsft (approved):

wwlorey (commented):
I think this is definitely on the right track. I do share your concern about how `getToken` may be used outside of the extension and I think it's worth keeping support for the "old way" of doing things in the legacy API. 

I put together some suggested changes that maintain compatibility in [this commit](https://github.com/microsoft/vscode-azure-account/commit/38ff35ba3003d54f0f63637edb9825edd39a532c):

* Change the current API to export `credentials2` as just type `TokenCredential` (no more `TokenCredentialsBase`). I don't think it's worth keeping around support for the deprecated typings for T1 in the current API. Especially since there's a big push to switch to the T2 SDK. `TokenCredentialsBase` is still included in the legacy API though.
* Rename `ForwardCompatibleToken` to `DeviceTokenCredentials2`. `DeviceTokenCredentialsX` has been the naming scheme for ADAL-related credentials and I vote we just use that name for the new class you created (it is just used with ADAL).
* Change the return type of `getToken` in `ForwardCompatibleToken` to `TokenResponse & AccessToken`. This ensures we're fully backwards/forwards compatible without making us rewrite `signRequest` and potentially having unforeseen consequences.

bwateratmsft:
> Change the return type of `getToken` in `ForwardCompatibleToken` to `TokenResponse & AccessToken`

Should we just make a single cross-compatible class, that can deal with either ADAL or MSAL underlying tokens, and implements `signRequest()` and `getToken(): TokenResponse & AccessToken`?

wwlorey:
> Should we just make a single cross-compatible class, that can deal with either ADAL or MSAL underlying tokens, and implements `signRequest()` and `getToken(): TokenResponse & AccessToken`?

I think it would be cleaner to keep them separated. The [`getToken` implementation for MSAL](https://github.com/microsoft/vscode-azure-account/blob/38ff35ba3003d54f0f63637edb9825edd39a532c/src/login/msal/PublicClientCredential.ts#L21) is more involved than for ADAL and I think it would end up being a big ole if/else in a shared class

nturinski:
> Change the return type of `getToken` in `ForwardCompatibleToken` to `TokenResponse & AccessToken`. This ensures we're fully backwards/forwards compatible without making us rewrite `signRequest` and potentially having unforeseen consequences.

This isn't possible due to the `TokenCredential.getToken` return type being only `TokenResponse` which is why I'm returning it as `<any>`.  This actually wouldn't work because we don't know when we should return a `TokenResponse` vs. when we should return a `AccessToken` so even if the return type was `TokenResponse & AccessToken`, it would only return an `AccessToken`

I have confidence that this approach would work for all our SDK packages because MSAL has been working without any issues, and it uses this implementation of `getToken`. The bigger concern is if there is anywhere `TokenCredential` calls `getToken`, it'll have the wrong token type.  I ran into this when I tried to use its `signRequest` command.

That being said, I only see `getToken` called in `signRequest` in `tokenCredentialsBase` (which `TokenCredential` extends) so we should be good.

bwateratmsft:
> This isn't possible due to the `TokenCredential.getToken` return type being only `TokenResponse`

Couldn't we could still mark the function return type as `any` to satisfy TypeScript, but the implementation could return an object that implements both?

nturinski:
Oh, good point.  None of the properties interfere with each other, so that should work!

alexweininger:
I think I understand this PR now, but for a while I was like:

wwlorey:
> we don't know when we should return a TokenResponse vs. when we should return a AccessToken

One thing to note is that the typing in our current API for `credentials2` doesn't explicitly include support for `signRequest`. That's something we're tacking on without alerting the public to - should we change that? Or should this be a workaround for our extensions that still rely on `signRequest`?

nturinski:
> > we don't know when we should return a TokenResponse vs. when we should return a AccessToken
> 
> One thing to note is that the typing in our current API for `credentials2` doesn't explicitly include support for `signRequest`. That's something we're tacking on without alerting the public to - should we change that? Or should this be a workaround for our extensions that still rely on `signRequest`?

I think I'm fine without exposing it.  It really isn't something we should be exposing since it's a bit of a workaround specific for us.

src/login/adal/DeviceTokenCredentials2.ts
```diff
@@ -0,0 +1,33 @@
...
+ * Token that is forward compatible with track 2 Azure SDK for Node.js
+ * `getToken: Promise<AccessToken>` is required for use with T2 Azure SDK, but doesn't
+ * affect T1 SDKs as those require `signRequest`
+ * `DeviceTokenCredentials` forces `getToken` to return `TokenResponse` so this is to
+ * overwrite that implementation
+ */
+export class DeviceTokenCredentials2 extends DeviceTokenCredentials implements TokenCredential {
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	public async getToken(): Promise<any> {
+		const tokenResponse = <AccessToken & TokenResponse>(await this.getTokenFromCache(this.username));
```
wwlorey, line 21: I think it's worth discussing this offline. I thought I understood the concerns brought up earlier in the thread but now I fear I may not 🧐

* Won't `tokenResponse` always just be type `TokenResponse` here? The type of the object after `Object.assign` is `AccessToken & TokenResponse` since we manually add the properties required for `AccessToken`. I know you brought up an issue with the return type here but I'm confused why it doesn't work with a `AccessToken & TokenResponse` return type?
* Is there a reason you're not using `super.getToken()` here?

src/login/adal/DeviceTokenCredentials2.ts
```diff
@@ -0,0 +1,33 @@
...
+		return Object.assign(tokenResponse, {token: tokenResponse.accessToken, expiresOnTimestamp: tokenResponse.expiresIn});
```
wwlorey, line 22: We could use `tokenResponse.expiresOn` here but that's a `Date | string` type so it seems more straightforward to do it this way

```suggestion
		return Object.assign(tokenResponse, {
			token: tokenResponse.accessToken, 
			expiresOnTimestamp: new Date().getTime() + tokenResponse.expiresIn 
		});
```

src/login/adal/DeviceTokenCredentials2.ts
```diff
@@ -0,0 +1,33 @@
...
+	}
+
+	public async signRequest(webResource: WebResource): Promise<WebResource> {
+		const tokenResponse: AccessToken = <AccessToken>(await this.getToken());
+			webResource.headers.set(
+				MSRestConstants.HeaderConstants.AUTHORIZATION,
+				`${MSRestConstants.HeaderConstants.AUTHORIZATION_SCHEME} ${tokenResponse.token}`
```
wwlorey, line 29: I don't think we need this function because we're only adding to the return type of `this.getToken()`, not removing anything that used to be there. If that's the case we can rely on the existing implementation of `signRequest` from the parent class. But like I said I think I may be misunderstanding something...

src/login/adal/DeviceTokenCredentials2.ts
nturinski: I must be going bonkers here.  Before, when I tried to have the return type be `AccessToken` it was throwing an error.  I guess I never tried changing the return type when I started returning `AccessToken & TokenResponse`.

And yeah, it'll always be `tokenResponse`.  Sorry for the confusion, I originally assigning the properties directly so I had to specify it was an `AccessToken` before I could set the properties, but then I changed it to an `Object.assign`

And the reason I'm not using `super.getToken` is because it didn't occur to me 😅

src/login/adal/DeviceTokenCredentials2.ts
nturinski: Good catch.  I misunderstood `expiresIn` to be the time it expired in as a ms number.

src/login/adal/DeviceTokenCredentials2.ts
nturinski: I think this will work now that we are implementing `TokenResponse & AccessToken`.  Originally, I only was returning an `AccessToken` so of course `signRequest` was failing because it uses `TokenResponse` to build the request.

nturinski:
> One thing to note is that the typing in our current API for `credentials2` doesn't explicitly include support for `signRequest`. That's something we're tacking on without alerting the public to - should we change that? Or should this be a workaround for our extensions that still rely on `signRequest`

I flipped on this.  We need to expose the new token for azuretools-extensionui since it uses the azure-account.api to set the type for credentials in `ISubscriptionContext`.

nturinski:
I'm going to go ahead and merge this, but of course, if there's any more feedback please feel free to comment on here or message me directly!

PR #423, opened 2022-02-04
Suppress credscan for suspected password found in extension.bundle.js.map

wwlorey:
I searched for the culprit for this credscan hit and was unable to find it. I argue that it's OK to suppress for the reasons listed in `_justification`.


bwateratmsft (approved):
I think we could suppress the `dist` folder entirely, none of that is truly source code.

PR #428, opened 2022-02-10
Mark `AzureSession` as deprecated in legacy API

wwlorey:


wwlorey:
`credentials` is no longer supported in the current extension API.

bwateratmsft (approved):

PR #431, opened 2022-02-10
Add new telemetry events

wwlorey:


src/login/updateSubscriptions.ts
```diff
@@ -100,5 +104,6 @@ async function loadSubscriptions(): Promise<AzureSubscription[]> {
 	const subscriptions: AzureSubscription[] = (<AzureSubscription[]>[]).concat(...lists);
 	// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
 	subscriptions.sort((a, b) => a.subscription.displayName!.localeCompare(b.subscription.displayName!));
+	context.telemetry.measurements.numSubscriptions = subscriptions.length;
```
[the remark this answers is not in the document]
bwateratmsft: Yeah, that's a good point--remembering to use `measurements` might end up wasting more time than the query optimization will fix.

src/extension.ts
```diff
@@ -38,9 +38,10 @@ export async function activateInternal(context: ExtensionContext, perfStats: { l
 	await callWithTelemetryAndErrorHandling('azure-account.activate', async (activateContext: IActionContext) => {
 		activateContext.telemetry.properties.isActivationEvent = 'true';
 		activateContext.telemetry.properties.activationTime = String((perfStats.loadEndTime - perfStats.loadStartTime) / 1000);
+		activateContext.telemetry.properties.authLibraryOnStartup = getSettingValue(authLibrarySetting) || 'undefined';
```
nturinski, line 41: I think you could just wrap this with a `String()`.  I don't have a strong preference, but just think it looks a bit cleaner.

wwlorey: I didn't think of that, nifty!

Issue #391, opened 2021-12-30
[Remote Codespaces] Pop up an error "The terminal process...failed to launch (exit code: 9)" when executing "Azure Cloud Shell (Bash)" 


v-mengwe:
**OS:** Remote Codespaces
**Build Version:** 20211228.1
**Regression:** Not a regression

**Repro Steps:**

1. Set the "Azure: Authentication Library" setting to "MSAL" -> Make sure you are signed out of Azure.
2. Open VS Code's Terminal Pane -> Click the dropdown -> Select "Azure Cloud Shell (Bash)".
3. Check whether there is an error message "Cloud Shell does not currently support authenticating with MSAL..." in the terminal.

**Expect:**
There is an error message "Cloud Shell does not currently support authenticating with MSAL..." in the terminal, like below:
![image](https://user-images.githubusercontent.com/62053665/147740788-3d71fe69-f26b-4f7d-9d64-205983875a7e.png)

**Actual:**
Pop up an error "The terminal process...failed to launch (exit code: 9)", like below:
![image](https://user-images.githubusercontent.com/62053665/147740893-cb221eb0-3c34-49d4-b6b8-398290461d76.png)

**More Info:**
This issue also reproduces when the "Azure: Authentication Library" setting is "ADAL".

dozer75:
I get the same error, but only when using Visual Studio Code against WSL and the Azure Account extension. Running Visual Studio Code and Azure Account extension without WSL works.

My reproduce steps are a bit simpler:
1. Install Azure Account extension
2. Log in using "Azure: Log in" command
3. Use the "Azure: Cloud Shell (Bash)" command

And I get the following error (the same way as above):
`The terminal process "/home/rg/.vscode-server/extensions/ms-vscode.azure-account-0.9.11/bin/node.sh '/home/rg/.vscode-server/bin/899d46d82c4c95423fb7e10e68eba52050e30ba3/node', '-e', 'require('/home/rg/.vscode-server/extensions/ms-vscode.azure-account-0.9.11/dist/cloudConsoleLauncher').main()', '--ms-enable-electron-run-as-node'" failed to launch (exit code: 9).`

Running Visual Studio Code without WSL and it works as expected.

Environment:

Operating system: `Windows 11 Pro OS Build 22000.376`

WSL Distro: `Ubuntu 20.04.3 LTS`

Visual Studio Code:  `1.63.2 899d46d82c4c95423fb7e10e68eba52050e30ba3 x64`

[...]

PR #432, opened 2022-02-10
Update when `ms-enable-electron-run-as-node` flag is added to cloud shell process

wwlorey:
Fixes https://github.com/microsoft/vscode-azure-account/issues/391

Codespaces running in web (Linux OS) *does not* need the flag while normal Linux running on desktop *does* need the flag. 

This gets messy because Codespaces running locally *does not* need the flag but as far as I can tell there isn't a way to know when the extension is being run from local Codespaces. VS Code's `appHost` and `uiKind` both read as `desktop` in this situation. 
So I'm making the conscious decision to not support cloud shell on local Codespaces because it doesn't seem worth the investigation time. With this change, we'll support cloud shell on Windows, Mac, Linux, and Codespaces Web.

For context, here's the flag requirements I gathered from my testing:

| Platform | Flag required? |
|-|:-:|
|Windows|❌|
|Mac|✅|
|Linux|✅|
|Codespaces (Linux) Web|❌|
|Codespaces (Linux) Local|❌|

And any time the flag is included where it isn't needed a big nasty error is shown (see above issue).

### DIFF
--- a/src/cloudConsole/cloudConsole.ts
+++ b/src/cloudConsole/cloudConsole.ts
@@ -273,7 +273,7 @@ export function createCloudConsole(api: AzureAccountExtensionApi, osName: OSName
 				shellArgs.shift();
 			}
 
-			if (process.platform === 'darwin' || process.platform === 'linux' && env.uiKind === UIKind.Desktop && semver.gte(version, '1.62.1')) {
+			if (!isWindows && env.uiKind === UIKind.Desktop && semver.gte(version, '1.62.1')) {
 				// https://github.com/microsoft/vscode/issues/136987
 				// This fix can't be applied to all versions of VS Code. An error is thrown in versions less than the one specified
 				shellArgs.push('--ms-enable-electron-run-as-node');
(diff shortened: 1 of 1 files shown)


bwateratmsft (approved):

ashdhalama:
I am experience a similar error message when try and open azure shell from wsl2(ubuntu-20.04).

The terminal process "/home/dhalama/.vscode-server/extensions/ms-vscode.azure-account-0.10.1/bin/node.sh '/home/dhalama/.vscode-server/bin/8908a9ca0f221f36507231afb39d2d8d1e182702/node', '-e', 'require('/home/dhalama/.vscode-server/extensions/ms-vscode.azure-account-0.10.1/dist/cloudConsoleLauncher').main()', '--ms-enable-electron-run-as-node'" failed to launch (exit code: 9).

wwlorey:
@ashdhalama That's expected behavior at this time. Local Codespaces is not supported and WSL likely falls into the same boat because they behave similarly.

From above:

> So I'm making the conscious decision to not support cloud shell on local Codespaces because it doesn't seem worth the investigation time. With this change, we'll support cloud shell on Windows, Mac, Linux, and Codespaces Web.

Please file an issue and we can reassess resources for fixing this problem.

Issue #433, opened 2022-02-11
Changing the auth library when the extension has yet to activate results in an error upon activation


v-mengwe:
**OS:** Win10/Mac
**Azure Account Build Version:** 20220210.28
**Azure Virtual Machine Build Version:** 20220208.7
**Regression:** Not a regression

**Repro steps:**

1. Make sure the "Azure: Authentication Library" setting is "ADAL".
2. Create a Linux VM with "Oracle Linux 8.4 (LVM)" image -> Right click the VM -> Connect to host via Remote SSH.
3. In the new opened VS Code window by the VM -> F1 -> Extensions: Install from VISX... -> Click "show Local" -> Select the visx file of Azure Account extension.
4. F1 -> Sign in -> Select an existing account.
5. Check whether there is no error.

**Expect:**
There is no error after signing in an existing account.

**Actual:**
There is an error after signing in an existing account.
![image](https://user-images.githubusercontent.com/62053665/153577499-1ee1290a-a3e9-4f1c-91c8-51cdf00aa41a.png)
Details of error:
```
Action: azure-account.login
Error type: SyntaxError
Error Message: Unexpected redacted:key in JSON at position 2
```

**More Info:**
The default value of "Azure: Authentication Library" setting is "MSAL" on the Remote setting after connecting successfully.

bwateratmsft:
I reproduced this by chance through another means, and the way of doing so isn't obvious, so I wanted to record it here.

1. Start with the ADAL auth mode. Sign in.
2. Reload the window. _Before_ the Azure extension activates, change the auth mode to MSAL.
3. Run command "Azure: Sign In" and follow through with the sign-in. Upon the callback to the extension's listener, this error is thrown.

bwateratmsft:
This issue could affect most customers that get assigned by the experimentation service into the treatment group (MSAL). Most likely, VSCode's experimentation client will alter the auth mode _before_ the Account extension has a chance to activate.

PR #434, opened 2022-02-11
Prompt to sign out and reload window if the auth library changed since last session

wwlorey:
Fixes https://github.com/microsoft/vscode-azure-account/issues/433

Thanks @bwateratmsft for the investigation you did on this one. 

This change adds caching for the last seen auth library setting. At startup, it compares that value to the current auth library and if they're different, the extension prompts the user to sign out & reload the window.

I also added the machine-overridable scope to the auth library setting as suggested in https://github.com/microsoft/vscode-azure-account/issues/433#issuecomment-1036348207.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -10,8 +10,9 @@ import { callWithTelemetryAndErrorHandling, createApiProvider, createAzExtOutput
 import { AzureExtensionApiProvider } from 'vscode-azureextensionui/api';
 import { AzureAccountExtensionApi } from './azure-account.api';
 import { createCloudConsole, OSes, OSName, shells } from './cloudConsole/cloudConsole';
-import { authLibrarySetting, cloudSetting, displayName, extensionPrefix, showSignedInEmailSetting } from './constants';
+import { AuthLibrary, authLibrarySetting, cloudSetting, displayName, extensionPrefix, showSignedInEmailSetting } from './constants';
 import { ext } from './extensionVariables';
+import { AuthLibraryCache, authLibraryCacheKey } from './login/AuthLibraryCache';
 import { AzureAccountLoginHelper } from './login/AzureLoginHelper';
 import { askForLogin } from './login/commands/askForLogin';
 import { loginToCloud } from './login/commands/loginToCloud';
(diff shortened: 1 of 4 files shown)


src/extension.ts
```diff
@@ -38,11 +39,12 @@ export async function activateInternal(context: ExtensionContext, perfStats: { l
 	await callWithTelemetryAndErrorHandling('azure-account.activate', async (activateContext: IActionContext) => {
 		activateContext.telemetry.properties.isActivationEvent = 'true';
 		activateContext.telemetry.properties.activationTime = String((perfStats.loadEndTime - perfStats.loadStartTime) / 1000);
-		activateContext.telemetry.properties.authLibraryOnStartup = String(getSettingValue(authLibrarySetting));
 
 		ext.experimentationService = await createExperimentationService(context);
 		ext.loginHelper = new AzureAccountLoginHelper(context, activateContext);
 
+		await checkAuthLibraryOnStartup(context, activateContext, ext.loginHelper);
```
wwlorey, line 46: I purposefully passed `ext.loginHelper` as a parameter here to enforce that class is instantiated before calling `checkAuthLibraryOnStartup`.

src/extension.ts
```diff
@@ -75,6 +77,46 @@ export async function activateInternal(context: ExtensionContext, perfStats: { l
 	return Object.assign(ext.loginHelper.legacyApi, createApiProvider([ext.loginHelper.api]));
 }
 
+async function checkAuthLibraryOnStartup(extensionContext: ExtensionContext, actionContext: IActionContext, loginHelper: AzureAccountLoginHelper): Promise<void> {
+	async function askThenSignOutAndReload(): Promise<void> {
+		const authLibraryChanged: string = localize('azure-account.authLibraryChanged', 'The authentication library has changed. Please sign out and reload the window for it to take effect.');
+		const signOutAndReload: string = localize('azure-account.signOutAndReload', 'Sign Out and Reload Window');
+		
+		// Purposefully await this message to block whatever command caused the extension to activate.
```
bwateratmsft, line 85: Alternatively, should we _not_ await this, and just throw a `UserCancelledError`? What happens if they just hit the X?

bwateratmsft (approved):

src/extension.ts
wwlorey: I vote we keep it as-is. I tried not awaiting this call and throwing a `UserCancelledError` immediately after and it continued on running the command anyway even after throwing the error.

How it currently is when we await the call: clicking X continues the command and results in the "Unexpected token..." error. But at that point I think it's the user's fault for ignoring our suggestion to sign out...

bwateratmsft: Ah, yeah you're right. The call to the command does not contain the above in the call stack so throwing the `UserCancelledError` will throw out of activation, not out of the command execution. Nevermind!

PR #437, opened 2022-02-14
Target VS Code Insiders for 0.10.0 release

wwlorey:


wwlorey:
I actually don't think this PR is doing what I was hoping it would. For example (based on [this comment](https://github.com/microsoft/vscode/issues/120274#issuecomment-936124933)): VS Code version 1.64.0 would satisfy the 1.63.0-insiders constraint. 

It's not clear to me that there's a way to only target insiders for the foreseeable future. What we could do is target the latest insiders version so that this release will be only pushed to insiders until the next stable version ships.

wwlorey:
Closing this. I didn't realize that we can scope the MSAL rollout to Insiders by using the experimentation framework

Issue #457, opened 2022-02-18
Exception when calling credentials2.getToken after upgrading to v0.10.0


sevillal:
- VS Code Version: 1.64.2
- Azure Account Extension Version: v0.10.0
- OS Version: Windows 11

Steps to Reproduce:

1. Open a folder in vscode
2. Start using extension version 0.9.11
2. Sign in
3. Upgrade to version 0.10.0
4. Using Azure ML Extension run command "Azure ML: Set Default Workspace" and select a default workspace
5. Reload the window
6. As part of the Azure ML Extension startup operations it will request credentials2.getToken on the subscription of the default workspace
7. We are seeing the following exception:
`[Error - 2022-02-17 21:44:53.314] TypeError: Cannot read property 'trim' of undefined
    at <REDACTED>.vscode\extensions\ms-vscode.azure-account-0.10.0\dist\extension.bundle.js:2:168982
    at Array.map (<anonymous>)
    at Function.trimArrayEntries (<REDACTED>\.vscode\extensions\ms-vscode.azure-account-0.10.0\dist\extension.bundle.js:2:168956)
    at new ScopeSet (<REDACTED>\.vscode\extensions\ms-vscode.azure-account-0.10.0\dist\extension.bundle.js:2:185274)
    at SilentFlowClient.<anonymous> (<REDACTED>\.vscode\extensions\ms-vscode.azure-account-0.10.0\dist\extension.bundle.js:2:247421)
    at step (<REDACTED>\.vscode\extensions\ms-vscode.azure-account-0.10.0\dist\extension.bundle.js:2:147853)
    at Object.next (<REDACTED>\.vscode\extensions\ms-vscode.azure-account-0.10.0\dist\extension.bundle.js:2:147958)
    at <REDACTED>\.vscode\extensions\ms-vscode.azure-account-0.10.0\dist\extension.bundle.js:2:146868
    at new Promise (<anonymous>)
    at __awaiter (<REDACTED>\.vscode\extensions\ms-vscode.azure-account-0.10.0\dist\extension.bundle.js:2:146554)
    at SilentFlowClient.acquireCachedToken (<REDACTED>\.vscode\extensions\ms-vscode.azure-account-0.10.0\dist\extension.bundle.js:2:246780)
    at SilentFlowClient.<anonymous> (<REDACTED>\.vscode\extensions\ms-vscode.azure-account-0.10.0\dist\extension.bundle.js:2:246481)
[...]

PR #462, opened 2022-02-18
Make `scopes` argument optional for MSAL's `getToken`

wwlorey:
Fixes https://github.com/microsoft/vscode-azure-account/issues/457 sort of.

This problem was related to MSAL and how `getToken` required a `scopes` parameter. Now `getToken` will successfully return an access token in this case. 

There's a problem with the new token though: it doesn't appear to have the correct scope(s) required for the ML extension to open a compute instance and this error is produced:

Screen Shot 2022-02-18 at 8 56 09 AM

It should be on the API users to hunt down the proper scopes required in the `getToken` call. We cannot & should not include all the scopes every extension needs in every token request. But finding the right scopes is a task that I wouldn't wish on my worst enemy. So we'll need to work with the MSAL team to figure out how to make that suck less for our API users.

### DIFF
--- a/src/login/msal/PublicClientCredential.ts
+++ b/src/login/msal/PublicClientCredential.ts
@@ -19,8 +19,12 @@ export class PublicClientCredential implements TokenCredential, AzExtServiceClie
 		this.accountInfo = accountInfo;
 	}
 
-	public async getToken(scopes: string | string[]): Promise<AccessToken | null> {
-		scopes = Array.isArray(scopes) ? scopes : [scopes];
+	public async getToken(scopes?: string | string[]): Promise<AccessToken | null> {
+		if (scopes) {
+			scopes = Array.isArray(scopes) ? scopes : [scopes];
+		} else {
+			scopes = [];
+		}
 
 		if (scopes.length === 1 && scopes[0] === 'https://management.azure.com/.default') {
 			// The Azure Functions & App Service APIs only accept the legacy scope (which is the default scope we use)


src/login/msal/PublicClientCredential.ts
```diff
@@ -19,8 +19,12 @@ export class PublicClientCredential implements TokenCredential, AzExtServiceClie
 		this.accountInfo = accountInfo;
 	}
 
-	public async getToken(scopes: string | string[]): Promise<AccessToken | null> {
-		scopes = Array.isArray(scopes) ? scopes : [scopes];
+	public async getToken(scopes?: string | string[]): Promise<AccessToken | null> {
+		if (scopes) {
+			scopes = Array.isArray(scopes) ? scopes : [scopes];
+		} else {
+			scopes = [];
```
bwateratmsft, line 26: If no scopes are passed in, do we want to use `defaultMsalScopes` instead of an empty array?
```suggestion
			scopes = defaultMsalScopes;
```
A token with no scopes is useless right? At least one with the default has a good chance of working (right?).

bwateratmsft (approved):

src/login/msal/PublicClientCredential.ts
wwlorey: Believe it or not providing no scopes actually gives you with a functional access token 🤷‍♂️ 

For instance in my testing with the ML extension I was able to list all resources in the tree view with an access token I obtained with `scopes = []`.

bwateratmsft: *sigh* of course it works. Uncut key opens all locks.
image

🤷‍♂️

PR #470, opened 2022-02-21
Include cloud instance and tenant in MSAL auth calls

wwlorey:
Fixes https://github.com/microsoft/vscode-azure-account/issues/441


wwlorey:
I verified these changes with default Azure, Azure China, & Azure Germany. CTI has access to China & Germany accounts for testing so I'll add test cases for those clouds.

bwateratmsft (approved):

PR #471, opened 2022-02-21
Tenant setting-related changes

wwlorey:
* Fix infinite loop in `loadTenants` when there's an error in the server response
* Finish running `Azure: Sign In to Azure Cloud` command even when the tenant ID is set to an empty string
* Get current target when updating setting value. This fixes the "Unable to write to Workspace Settings because no workspace is opened. Please open a workspace first and try again." error while running `Azure: Select Tenant` command

### DIFF
--- a/src/login/commands/loginToCloud.ts
+++ b/src/login/commands/loginToCloud.ts
@@ -50,7 +50,7 @@ export async function loginToCloud(): Promise<void> {
 				prompt: localize('azure-account.enterTenantId', "Enter the tenant id"),
 				placeHolder: localize('azure-account.tenantIdPlaceholder', "Enter your tenant id, or '{0}' for the default tenant", commonTenantId),
 				ignoreFocusOut: true});
-			if (tenantId) {
+			if (tenantId !== undefined) {
 				if (armUrl) {
 					await config.update(customCloudArmUrlSetting, armUrl, getCurrentTarget(config.inspect(customCloudArmUrlSetting)));
 				}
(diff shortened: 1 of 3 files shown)


src/login/commands/loginToCloud.ts
```diff
@@ -50,7 +50,7 @@ export async function loginToCloud(): Promise<void> {
 				prompt: localize('azure-account.enterTenantId', "Enter the tenant id"),
 				placeHolder: localize('azure-account.tenantIdPlaceholder', "Enter your tenant id, or '{0}' for the default tenant", commonTenantId),
 				ignoreFocusOut: true});
-			if (tenantId) {
+			if (tenantId !== undefined) {
```
bwateratmsft, line 53: What happens if they submit an empty string? Previously that would have fallen into the `else`, but now it will go into the `if`.

bwateratmsft (approved):

src/login/commands/loginToCloud.ts
wwlorey: Given an empty string for the tenant ID: this logic used to stop the `Azure: Sign In to Azure Cloud` command in its tracks and not save either the newly selected cloud or the new tenant ID to settings. This wasn't ideal because an empty string for the tenant setting is both a common and acceptable value. 

There's logic everywhere we read the tenant setting that defaults to use the "common" tenant in case that setting is empty/undefined ([here](https://github.com/microsoft/vscode-azure-account/blob/abd2f66635549ff35ff87140398020c28ed546ce/src/login/AzureLoginHelper.ts#L126) and [here](https://github.com/microsoft/vscode-azure-account/blob/abd2f66635549ff35ff87140398020c28ed546ce/src/login/AzureLoginHelper.ts#L176)).

PR #472, opened 2022-02-21
Changes to login and initialization

wwlorey:
When testing https://github.com/microsoft/vscode-azure-account/pull/470 I noticed that switching between sovereign clouds while also changing the tenant using the `Azure: Sign In to Azure Cloud` command resulted in several errors ranging from cache read errors to "invalid grant".

This was happening because [this config watcher]( https://github.com/microsoft/vscode-azure-account/blob/36e0a4727655637eed6dd4ea84503e290f1fe660/src/login/AzureLoginHelper.ts#L74-L79) was activating twice: once for the cloud change and again for the tenant change. This was on top of the initialization function being called upon extension activation. These calls resulted in the cache being in a strange state when `loginSilent` was called and caused the above errors.

To fix this I force a logout before every login. This doesn't affect the API (it still shows the "signing in..." status) and it ensures the cache is clean before attempting to login with a new configuration.

### DIFF
--- a/src/login/AzureLoginHelper.ts
+++ b/src/login/AzureLoginHelper.ts
@@ -7,7 +7,7 @@ import { Environment } from '@azure/ms-rest-azure-env';
 import { callWithTelemetryAndErrorHandling, IActionContext } from '@microsoft/vscode-azext-utils';
 import { CancellationTokenSource, commands, EventEmitter, ExtensionContext, MessageItem, window, workspace } from 'vscode';
 import { AzureLoginStatus, AzureResourceFilter, AzureSession, AzureSubscription } from '../azure-account.api';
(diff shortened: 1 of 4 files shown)


src/login/AzureLoginHelper.ts
```diff
@@ -170,7 +165,7 @@ export class AzureAccountLoginHelper {
 		this.updateLoginStatus();
 	}
 
-	private async initialize(trigger: LoginTrigger, doLogin?: boolean, migrateToken?: boolean): Promise<void> {
+	private async initialize(trigger: LoginTrigger): Promise<void> {
```
wwlorey, line 168: Removing the `initialize` call in `onDidChangeConfiguration` meant that these arguments were no longer needed along with a few of the `LoginTrigger` strings.

bwateratmsft (approved):

Issue #468, opened 2022-02-20
Add a mechanism for cancelling the login operation


tst32:
it is not an important bug, just want an option to cancel gracefully login request
Does this occur consistently? Yes
Repro steps:

1. When do ocassionaly pressed azure func shortcut in vscode
2. i don't wnat to login but want to close azure func i have no option bu wait for timeout error

Action: azure-account.login
Error type: Error
Error Message: Timeout waiting for code

Version: 0.10.0
OS: linux
OS Release: 5.4.0-100-generic
Product: Visual Studio Code
Product Version: 1.64.2
Language: en

<details>
<summary>Call Stack</summary>

```
Timeout.<anonymous> extension.bundle.js:2:2321688
listOnTimeout timers.js:554:17
processTimers timers.js:497:7
```

</details>

wwlorey:
Hi @tst32, does running the `Azure: Sign Out` command via the [command palette](https://code.visualstudio.com/docs/getstarted/userinterface#_command-palette) accomplish this for you?

tst32:
![image](https://user-images.githubusercontent.com/23115742/155007241-49eece03-cd66-429b-8139-ce6894f77fa8.png)
No, <kbd> Cntrl </kbd> <kbd> Shift </kbd> <kbd> P </kbd> `Azure sign out` does nothing

wwlorey:
Let's add a cancellable, long-running notification that's shown for the duration of the sign in process.

wwlorey:
@tst32 FYI we released this feature in [v0.10.1](https://github.com/microsoft/vscode-azure-account/releases/tag/v0.10.1).

tst32:
well implemented!

PR #480, opened 2022-02-24
Show cancellable, long running notification when signing in

wwlorey:
Fixes https://github.com/microsoft/vscode-azure-account/issues/468

### DIFF
--- a/src/login/AuthProviderBase.ts
+++ b/src/login/AuthProviderBase.ts
@@ -55,7 +55,7 @@ export abstract class AuthProviderBase<TLoginResult> {
 			server.close(error => error && ext.outputChannel.appendLog(parseError(error).message));
 			clearTimeout(codeTimer);
 			context.telemetry.properties.serverClosed = 'true';
-			ext.outputChannel.appendLog(localize('azure-account.serverClosed', 'Server closed.'));
+			ext.outputChannel.appendLog(localize('azure-account.authProcessCancelled', 'Authentication process cancelled.'));
 		});
 
 		if (isAdfs) {
--- a/src/login/AuthProviderBase.ts
+++ b/src/login/AuthProviderBase.ts
@@ -6,13 +6,14 @@
 import { TokenCredential } from "@azure/core-auth";
 import { Environment } from "@azure/ms-rest-azure-env";
 import { AccountInfo } from "@azure/msal-node";
-import { parseError } from "@microsoft/vscode-azext-utils";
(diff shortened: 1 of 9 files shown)


src/login/AuthProviderBase.ts
```diff
@@ -48,7 +49,14 @@ export abstract class AuthProviderBase<TLoginResult> {
 		}
 
 		const nonce: string = randomBytes(16).toString('base64');
-		const { server, redirectPromise, codePromise } = createServer(nonce);
+		const { server, redirectPromise, codePromise, codeTimer } = createServer(context, nonce);
+
+		cancellationToken.onCancellationRequested(() => {
+			server.close();
```
alexweininger, line 55: According to https://nodejs.org/api/net.html#serverclosecallback `.close()` is async, and takes a callback. I think it'd be safest to use this callback in case it takes a while to close or there is an error.

alexweininger: Not sure, maybe we don't care 😅

wwlorey: The callback argument is optional and it's safe to omit it but I do think it's worth tracking any errors that come up. We don't expect this call to fail and if it does we don't really care (because we don't need the server at that point anyway) but I think it's better to know about the error than not.

alexweininger (approved):

src/login/AuthProviderBase.ts
```diff
@@ -48,7 +49,14 @@ export abstract class AuthProviderBase<TLoginResult> {
...
 
 		const nonce: string = randomBytes(16).toString('base64');
-		const { server, redirectPromise, codePromise } = createServer(nonce);
+		const { server, redirectPromise, codePromise, codeTimer } = createServer(context, nonce);
+
+		cancellationToken.onCancellationRequested(() => {
+			server.close(error => error && ext.outputChannel.appendLog(parseError(error).message));
+			clearTimeout(codeTimer);
+			context.telemetry.properties.serverClosed = 'true';
+			ext.outputChannel.appendLog(localize('azure-account.serverClosed', 'Server closed.'));
```
bwateratmsft, line 58: I don't think the average user will know what "server closed" means, and it might lead to confusion (e.g. "Is Microsoft's server closed?"). Might be better to just say that the auth process has been cancelled.

bwateratmsft (approved):

bwateratmsft (approved):

Issue #510, opened 2022-03-08
Update extension sample to use the versioned API


wwlorey:
From the changelog: 
> The extension now supports a [versioned API](https://github.com/microsoft/vscode-azure-account/blob/main/src/azure-account.api.d.ts) accessible by calling [getApi](https://github.com/microsoft/vscode-azuretools/blob/d38498f0085deb912675e4d2cb376f973c12f31e/utils/api.d.ts#L22) on the extension's exports.

wwlorey:
We should also consider adding a doc explaining how to transition to the new API for projects that are still on the legacy api

PR #517, opened 2022-03-16
Update sample extension

wwlorey:
* Use `getApi` from `@microsoft/vscode-azext-utils` to use the latest extension API instead of the legacy API
* Fix ES Lint errors
* Update a few packages to match the main extension package.json

Fixes https://github.com/microsoft/vscode-azure-account/issues/510

### DIFF
--- a/sample/package.json
+++ b/sample/package.json
@@ -45,7 +45,7 @@
   "devDependencies": {
     "@microsoft/vscode-azext-utils": "^0.1.0",
     "@types/node": "^14.0.0",
-    "@types/vscode": "^1.53.0",
+    "@types/vscode": "1.58.0",
     "typescript": "^4.0.2"
   },
   "dependencies": {
--- a/sample/package.json
+++ b/sample/package.json
@@ -5,10 +5,10 @@
   "repository": {
     "url": "https://github.com/microsoft/vscode-azure-account.git"
   },
-  "version": "0.1.0",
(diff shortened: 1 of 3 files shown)


sample/package.json
```diff
@@ -43,13 +43,14 @@
...
+    "@types/node": "^14.0.0",
+    "@types/vscode": "^1.53.0",
     "typescript": "^4.0.2"
   },
   "dependencies": {
     "@azure/arm-appservice": "^9.0.0",
     "@azure/arm-resources": "^4.2.2",
-    "@azure/arm-subscriptions": "^3.1.2"
+    "@azure/arm-subscriptions": "^3.1.2",
+    "@microsoft/vscode-azext-utils": "^0.1.0"
```
bwateratmsft, line 54: Is it possible for extensions to use the API without taking a dependency on this package?

wwlorey: We need some sort of dependency on `@microsoft/vscode-azext-utils` so that `getApi` can be used to access a versioned API. Otherwise people would fall back to using the legacy API which I'd like to avoid.

The Account Extension has a hard dependency on `@microsoft/vscode-azext-utils` so we can specify it as a dev dependency in the sample and it will have access to that package at runtime.

bwateratmsft (dismissed):

bwateratmsft (approved):

Issue #419, opened 2022-02-01
Help users with proxy/internet access related issues debug their connection


wwlorey:
I heard from a customer today that ran into an issue similar to https://github.com/microsoft/vscode-azure-account/issues/227. The difference was they were behind a corporate proxy - not using WSL.

It would have helped their situation if we provided a way to:
* Turn off the "is this machine online" check before we start the login process. Maybe through a VS Code setting. [^1]
* Know which URLs the extension is attempting to hit so the user can configure their proxy accordingly. This could be done in the error notifications or in the logs.

[^1]: This may not be the most useful addition because:
    1. It isn't super discoverable 
    2. The user is still required to whitelist domains even after skipping the online check

bwateratmsft:
What's the purpose of the online check?

wwlorey:
> What's the purpose of the online check?

Before the extension gets too deep into the login flow it pings the `activeDirectoryEndpointUrl` to make sure it has internet access. If not, it errors out early. That URL is then what's visited later on in the login process to hit the auth server and perform the login.

TimMarshAU:
I'm having loads of issues behind a Bluecoat proxy. Adding the PAC file doesn't help, nor does override. I'm also on a VPN (AOVPN Windows 10 Enterprise). Weirdly, was working yesterday, and as soon as I installed Node.js this AM it coincidentally stopped working. Simply cannot login to our Azure SMC. I can get VS to authenticate with a device code but VS Code sits there "signing in" for a lengthy time and without timeout.

wwlorey:
Hi @TimMarshAU, please file an issue on the extension with more details about your problem after following the steps in our [troubleshooting guide](https://github.com/microsoft/vscode-azure-account/wiki/Troubleshooting#unable-to-sign-in-while-using-a-proxy). It sounds like you've already tried the "Sig In with Device Code" workaround but there's another tip on that guide that may be helpful.

TimMarshAU:
Quick update: we found our proxy (Bluecoat) was intercepting all traffic to inspect which SSL doesn't like. Once we had the below URLs being tunnelled through the proxy, problem solved, including retrieving Azure subscriptions inside VS Code, and publishing Function App from VS Code.

https://management.azure.com/*
https://functionscdn.azureedge.net/*
https://vscode-redirect.azurewebsites.net/*
https://login.microsoftonline.com/*

PR #520, opened 2022-03-16
Log all URLs that the extension tries to reach

wwlorey:
Fixes https://github.com/microsoft/vscode-azure-account/issues/419

This PR specifically address this bullet from the above issue:

> Know which URLs the extension is attempting to hit so the user can configure their proxy accordingly. This could be done in the error notifications or in the logs.

I don't want to add a mechanism to turn off the "is this machine online" check for reasons listed in that issue.

I did my best to catch all the URLs and avoid too many duplicate URLs 🕵️‍♂️

Here's how the output channel looks during the sign in process:

Screen Shot 2022-03-16 at 2 58 21 PM


alexweininger (approved):
This is cool, I hope it proves helpful! It might be a good thing to be able to toggle on/off via settings. Ex: a "Debug my proxy" mode. But that might be overcomplicating a nice and simple feature like this.

bwateratmsft (approved):

wwlorey:
Discussed offline: we'll leave this feature as is without putting the logs behind a flag in settings.

Issue #380, opened 2021-12-03
The opened "Azure Cloud Shell (Bash)" terminal looks a little ugly after switching Authentication Library


v-ruizh:
**OS:** All
**Build Version:** 20211130.7
**Regression:** Not a regression

**Repro Steps:**
1. Set the "Azure: Authentication Library" setting to "ADAL".
2. Open a "Azure Cloud Shell (Bash)" terminal.
3. Change the "Azure: Authentication Library" setting to "MSAL"
4. Click "Sign Out and Reload Window" button in the opened notification to take effect Authentication Library change.
5. Check the above opened "Azure Cloud Shell (Bash)" terminal.

**Expect:**
The opened "Azure Cloud Shell (Bash)" terminal looks well.

**Actual:**
The opened "Azure Cloud Shell (Bash)" terminal looks a little ugly.
![image](https://user-images.githubusercontent.com/45614674/144560345-8a9a04d7-70e0-474d-bbba-4047c0c98d6a.png)

wwlorey:
Once VS Code supports opting out of persistent terminals we can fix this. We would just mark all of our terminals as non-persistent so when the window is refreshed that terminal is no longer shown. Tracking in https://github.com/microsoft/vscode/issues/118726

wwlorey:
This is coming soon: https://github.com/microsoft/vscode/issues/139112

PR #527, opened 2022-03-23
Add `isTransient` flag to cloud shell terminal

wwlorey:
Fixes https://github.com/microsoft/vscode-azure-account/issues/380

Support for this flag was introduced in VS Code v1.65.0 so I bumped types/vscode & the engine. It's a pretty aggressive version bump on our end though so if this change doesn't seem worth it I'm fine waiting on it.

### DIFF
--- a/src/cloudConsole/cloudConsole.ts
+++ b/src/cloudConsole/cloudConsole.ts
@@ -288,7 +288,8 @@ export function createCloudConsole(api: AzureAccountExtensionApi, osName: OSName
 				shellArgs,
 				env: {
 					CLOUD_CONSOLE_IPC: server.ipcHandlePath,
-				}
+				},
+				isTransient: true
 			};
 
 			const cleanupCloudShell = () => {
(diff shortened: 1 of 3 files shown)


src/cloudConsole/cloudConsole.ts
```diff
@@ -288,7 +288,8 @@ export function createCloudConsole(api: AzureAccountExtensionApi, osName: OSName
 				shellArgs,
 				env: {
 					CLOUD_CONSOLE_IPC: server.ipcHandlePath,
-				}
+				},
+				isTransient: true
```
bwateratmsft, line 292: What does this flag do?

bwateratmsft (approved):

src/cloudConsole/cloudConsole.ts
wwlorey: It stops the terminal from persisting on reload so we don't get ugly errors like this one:

image

Issue #511, opened 2022-03-09
Show toast notification prompting to sign out/back in when changing settings


wwlorey:
*This issue comes from an internal person who was confused when changing their Azure cloud setting didn't automatically change the cloud they were signed in with.*

We should watch all relevant settings (cloud, tenant, auth library) and throw this notification when they're changed. This applies when the extension is activated and settings watchers are alive.

We should also keep a cache of previously seen settings and if what we see on startup doesn't match the last seen setting: show this notification.

wwlorey:
This is very similar to the auth library on startup check, just with more settings supported (and additional settings watchers).

https://github.com/microsoft/vscode-azure-account/blob/4b10f7773c0338e72ddb054a6aa7a08e16b294bd/src/extension.ts#L96-L118

PR #562, opened 2022-05-04
Watch for settings changes and notify when a sign out/window reload is required

wwlorey:
Fixes https://github.com/microsoft/vscode-azure-account/issues/511


src/login/AzureLoginHelper.ts
```diff
@@ -73,21 +74,27 @@ export class AzureAccountLoginHelper {
...
-				await this.context.globalState.update(authLibraryCacheKey, { lastUsedAuthLibrary: newAuthLibrary });
-
-				const mustSignOutAndReload: string = localize('azure-account.mustSignOutAndReload', 'You must sign out and reload the window to authenticate with "{0}"', newAuthLibrary);
-				const signOutAndReload: string = localize('azure-account.signOutAndReload', 'Sign Out and Reload Window');
-				void window.showInformationMessage(mustSignOutAndReload, signOutAndReload).then(async value => {
-					if (value === signOutAndReload) {
-						actionContext.telemetry.properties.signOutAndReload = 'true';
-						await this.logout(true /* forceLogout */);
-						await commands.executeCommand('workbench.action.reloadWindow');
+			} else if (e.affectsConfiguration(getSettingWithPrefix(authLibrarySetting)) || e.affectsConfiguration(getSettingWithPrefix(cloudSetting)) || e.affectsConfiguration(getSettingWithPrefix(tenantSetting))) {
```
bwateratmsft, line 77: I'd alter this to check for anything within `cachedSettingKeys`. Even though we check each of the three individually on lines 87 and 93, since the purpose of this outer guard is to catch everything in that `cachedSettingKeys` array, I think it's best to do that.

e.g.
```suggestion
			} else if (cachedSettingKeys.some(k => e.affectsConfiguration(getSettingWithPrefix(k)))) {
```

src/login/checkSettingsOnStartup.ts
```diff
@@ -0,0 +1,84 @@
...
+import { commands, ExtensionContext, window } from "vscode";
+import { localize } from "../utils/localize";
+import { getSettingValue } from "../utils/settingUtils";
+import { AzureAccountLoginHelper } from "./AzureLoginHelper";
+import { cachedSettingKeys, SettingsCache, settingsCacheKey, SettingsCacheVerified } from "./SettingsCache";
+
+export async function checkSettingsOnStartup(extensionContext: ExtensionContext, actionContext: IActionContext, loginHelper: AzureAccountLoginHelper): Promise<void> {
+	const numSettings = cachedSettingKeys.length;
+
+	const lastSeenSettingsCache: SettingsCache | undefined = extensionContext.globalState.get(settingsCacheKey);
```
bwateratmsft, line 16: nit: `    ` instead of `\t`

src/login/checkSettingsOnStartup.ts
```diff
@@ -0,0 +1,84 @@
...
+    const valuesToCopy = lastSeenSettingsCache?.values?.length === numSettings ? lastSeenSettingsCache.values : [];
+    const resetLastSeenSettingsCache: boolean = valuesToCopy.length === numSettings ? false : true;
```
bwateratmsft, line 18: What you have is valid but another style that might be clearer:
```suggestion
    const resetLastSeenSettingsCache: boolean = valuesToCopy.length !== numSettings;
```

src/login/checkSettingsOnStartup.ts
```diff
@@ -0,0 +1,84 @@
...
+    actionContext.telemetry.properties.resetLastSeenSettingsCache = String(resetLastSeenSettingsCache);
+    const lastSeenSettingsCacheVerified: SettingsCacheVerified = { values: new Array<undefined>(numSettings) };
+    lastSeenSettingsCacheVerified.values.splice(0, numSettings, ...valuesToCopy)
+
+	const lastSeenSettingsCacheNew: SettingsCacheVerified = { values: new Array<string | undefined>(numSettings) };
+
+    // Ask to sign out & reload if the cache is being reset.
+    let shouldSignOutAndReload: boolean = resetLastSeenSettingsCache;
+
+	for (let index = 0; index < numSettings; index++) {
```
bwateratmsft, line 28: nit: more switching between tabs and spaces

src/login/checkSettingsOnStartup.ts
```diff
@@ -0,0 +1,84 @@
...
+    // Purposefully await this message to block whatever command caused the extension to activate.
+    await window.showInformationMessage(signInAgainRequired, signIn).then(async value => {
+        if (value === signIn) {
+            actionContext.telemetry.properties.signInAgain = 'true';
+            await commands.executeCommand('azure-account.login');
+        }
+    });
+}
+
+function sendTelemetryEvent(actionContext: IActionContext, eventPrefix: string, eventSuffix: string, value: string | undefined): void {
```
bwateratmsft, line 82: Maybe this? We aren't actually sending the event in this method.

```suggestion
function addPropertyToTelemetry(actionContext: IActionContext, eventPrefix: string, eventSuffix: string, value: string | undefined): void {
```

src/login/checkSettingsOnStartup.ts
wwlorey: Good catch - I turned my auto formatter off a while ago because it was causing problems. And yet there are still problems 😄

bwateratmsft (dismissed):

PR #624, opened 2022-08-29
Add Support for Workspace Trust

tecton:
Azure account extension will be activated in "Restricted Mode" with limited support.

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,17 +90,23 @@
             }
         ],
         "menus": {
+            "commandPalette": [
+                {
+                  "command": "azure-account.uploadFileCloudConsole",
+                  "when": "isWorkspaceTrusted"
+                }
+            ],
             "explorer/context": [
                 {
                     "command": "azure-account.uploadFileCloudConsole",
-                    "when": "resourceScheme == file && !explorerResourceIsFolder && openCloudConsoleCount && openCloudConsoleCount != 0",
+                    "when": "resourceScheme == file && !explorerResourceIsFolder && isWorkspaceTrusted && openCloudConsoleCount && openCloudConsoleCount != 0",
                     "group": "999_cloudConsole"
                 }
             ],
             "editor/title/context": [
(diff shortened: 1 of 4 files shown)


bwateratmsft (commented):
This looks good to me. There are just a couple of changes I'd like to add though, to make the `azure-account.uploadFileCloudConsole` command require trust:

- In the `"menus"` section, add `&& isWorkspaceTrusted` to the `"when"` clause contexts for the `azure-account.uploadFileCloudConsole` command in both `"explorer/context"` and `"editor/title/context"`.
- Add a new section to `"menus"` that also guards the `azure-account.uploadFileCloudConsole` command behind `isWorkspaceTrusted`.
- Add a check in the code for `azure-account.uploadFileCloudConsole` to `vscode.workspace.isTrusted` and throw if it is not. Strictly speaking this shouldn't be needed--it's just defense in depth--but the documentation around workspace trust does recommend it.

bwateratmsft:
Additionally, @philliphoff should we require workspace trust to launch a cloud console?

bwateratmsft (approved):
This looks great to me, thanks @tecton! Can you also take a look @philliphoff?

alexweininger (dismissed):

philliphoff:
> Additionally, @philliphoff should we require workspace trust to launch a cloud console?

I can imagine a similar set of concerns for that command as with file upload, so I think that'd make sense.  That said, I'm fine with that being a follow up item.

philliphoff (dismissed):

tecton:
Hi @philliphoff @bwateratmsft @alexweininger, I added a check when opening a Cloud console, it will close the terminal and show a notification if it's not in a trusted workspace. It's also defense in depth because VS Code will block the opening of a terminal in "Restricted Mode" so that users will not be able to open the Cloud console.

philliphoff (approved):

PR #712, opened 2023-02-02
Configure global-agent

alexweininger:
Another possible solution to proxy issues. 

https://www.npmjs.com/package/global-agent


src/utils/configureGlobalAgent.ts
```diff
@@ -0,0 +1,18 @@
...
+
+import { bootstrap } from "global-agent";
+import { workspace } from "vscode";
+
+export async function configureGlobalAgent(): Promise<void> {
+
+    workspace.getConfiguration('http.proxy')
+
+    // lowercase environment variables take precedence over uppercase ones
+    process.env.GLOBAL_AGENT_HTTPS_PROXY = workspace.getConfiguration('http').get('proxy') || process.env.https_proxy || process.env.HTTPS_PROXY || process.env.http_proxy || process.env.HTTP_PROXY;
```
bwateratmsft, line 14: Is this a standard order or is it something we should try to document somewhere?

bwateratmsft (dismissed):

src/utils/configureGlobalAgent.ts
alexweininger: It's not standard, I followed the order VS Code used to configure their proxy.

bwateratmsft: Good idea, a link to the source or doc or whatever indicating that order is helpful.

bwateratmsft (approved):

bwateratmsft (approved):

PR #851, opened 2023-08-29
Remove use of AAD Graph for cloud shell

### LINKED ISSUE #849: Migrate from AAD graph to MS graph
AAD graph (graph.windows.net) will be retired on or after 30 September without notice. Need to migrate to MS graph (graph.microsoft.com).

alexweininger:
Closes #849 

AAD Graph was being used by cloud shell in two ways:

1. If a user had subscriptions from multiple tenants, we let the user select which tenant to launch the cloud shell in. A Graph API endpoint for listing user tenants was used to populate the quick pick.

    I replaced this with a call to the new List Tenants ARM endpoint. https://learn.microsoft.com/en-us/rest/api/resources/tenants/list

2. We fetched an AAD Graph token and passed it to Cloud Shell upon initialization.

    I just stopped fetching and passing this token. My reasoning is that any features relying on this token will also break once AAD Graph is gone. So there's no reason to keep passing this token.

### DIFF
--- a/src/cloudConsole/cloudConsole.ts
+++ b/src/cloudConsole/cloudConsole.ts
@@ -9,7 +9,6 @@ import * as FormData from 'form-data';
 import { ReadStream } from 'fs';
 import * as http from 'http';
 import { ClientRequest } from 'http';
-import { DeviceTokenCredentials } from 'ms-rest-azure';
 import { Socket } from 'net';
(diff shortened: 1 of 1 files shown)


bwateratmsft (approved):

alexweininger:
Waiting on some confirmation from the cloud shell team on these changes before merging.

PR #958, opened 2024-05-07
Depend on Azure Resources extension for Cloud Shell feature

alexweininger:
Related to https://github.com/microsoft/vscode-azureresourcegroups/pull/848

The idea is to make RGs an extension dependency so that users don't notice any UI changes. However, the entry point for cloud shell will now be contributed by RGs instead of Azure Account. This will be a great way to easily transition users to RGs.

Keeping the cloud shell code around because some extensions still rely on the API, which has yet to be migrated to RGs. One of which is https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-azureterraform

They can rely on AA until we figure out a solution

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -6,7 +6,7 @@
     },
     "typescript.tsdk": "./node_modules/typescript/lib", // we want to use the TS server from our node_modules folder to control its version
     "editor.codeActionsOnSave": {
-        "source.fixAll.eslint": true,
-        "source.organizeImports": true
+        "source.fixAll.eslint": "explicit",
+        "source.organizeImports": "explicit"
     }
 }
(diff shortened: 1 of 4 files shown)


MicroFish91 (approved):

nturinski (approved):

nturinski:
Offline: Told Alex that I think that it was fine to not remove all the unused code since that could potentially cause more headaches than just removing the entry point does.

alexweininger:
> Worth trying out--what if someone installs this version of the Azure Account extension (that depends on Resources), plus an older version of the Azure Resources extension that depends on Azure Account. Circular dependency. Is there an easy escape?

I tried this out and 1. it was really hard to do this, and 2. I could fix it by uninstalling all of the Azure extensions and then installing the new ones.

